PR #33, opened 2021-02-12
Use box-shadow instead of border

msal4:
Using `box-shadow` doesn't take space


vhoyer:
NICE MAN, I was getting triggered by the jumping profile haha

PR #77, opened 2021-02-12
Socket handler refactor

Qizot:
This is a small refactor of socket handler. I wanted @benawad to give you a hint how elixir code can be utilized to be more readable. Remember this couple of things:
- pattern matching is super fast and you can literally do magic with it
- `{:ok, data}, {:error, reason}` tuples are your best friends
- use aliases whenever possible if you happen to use some functions more than once
- if you see that your function gets too big (> 50 lines) try to break it into several smaller functions
- `|>` operator makes your code more readable and functional, try to compose functions with it as much as possible
- elixir style does not really like camel case variables, you can keep them in jsons but it's a standard to use snake case as much as possible
- it's better to write some function for your genservers modules that take as a first argument the genserver process rather than calling raw `GenServer.call` and `GenServer.cast`

And the most crucial thing: `typespecs`. I saw you complaining on your recent video with elixir that you couldn't get autocomplete to work (or that it does not exist). If you  do the right `@spec` for functions and `@type` for your structs or types necessary for function specs you can greatly benefit from it. Dialyzer will tell you when you wrongly call your functions and what are the expected types. You can have mouse-hover information about what function expects and what it returns. And when your cursor is inside of a struct you get the full auto-complete of existing fields.

One more advice, look up how configs in elixir work. You should not have runtime-loaded variables inside `prod/test/dev` configs but rather move that to `runtime.exs` or `releases.exs` where the prior is preferable.
If you require env variables in your dev config then nobody is even able to run `mix format` on it.

You mentioned your relation with elixir as love-hate, I hope that it will go more towards love when you get to know it better :)

Wish you all the best and keep the great youtube work.


benawad:
This is amazing!!!

I'm excited to start using this everywhere, thank you :)

Issue #76, opened 2021-02-12
show button/icon function on hover


Veekthor:
I had to click the buttons at the bottom of the chat room to know what they do. I ended up being a speaker at some point. A tooltip on hover would be nice :-)

kweken:
Yes please do this one, had the same happen to me

Issue #65, opened 2021-02-12
Button Text Transform


MikeSundaysProgrammer:
This is a minor issue, but I think the UI could be improved if the buttons have the text in all caps like most websites have

Issue #49, opened 2021-02-12
Labels for buttons


tomowen02:
I think it would be useful to have labels for when you hover your mouse above a button. Just means you don't have to guess what things mean

mohamed-azhar:
Tooltips!

PR #82, opened 2021-02-12
Add button popovers with title attribute

KieranHolroyd:


KieranHolroyd:
fixes #76

benawad:
Doesn't the browser tooltip suck, shouldn't we use a tooltip library?

leandrasilver (changes requested):
Looks good in terms of this issue raised.

Since this is a web app, and potentially looking for investors, we also need to think in terms of accessibility. Keyboard users will not be able to view the title attribute on buttons. As for screenreader users please, can you also add `aria label` attribute so users will also get this information, since the button itself has no text.

KieranHolroyd:
tooltip libraries are kinda hit or miss imo, it's easy enough to implement one after the fact, +1 for the accessibility idea

KieranHolroyd:
also if i'm honest, i don't wanna break things and i can't test the frontend without starting elixir, and i have no idea how to do that yet sorry

MrMalloc:
What do you think about downsizing the svgs to 24x24 and adding a text below the icon.
Something like https://material.io/components/bottom-navigation
This would also solve #65 and #49

leandrasilver:
> What do you think about downsizing the svgs to 24x24 and adding a text below the icon.
> Something like https://material.io/components/bottom-navigation
> This would also solve #65 and #49

+1

PR #98, opened 2021-02-13
Refactor this redundant 'await' on a non-promise.

aofdev:
This parameter "**handler**" on values that are not Promise.

It is possible to use await on values that are not Promises, but it's useless and misleading.


shawarma/src/utils/startRabbit.ts
```diff
@@ -49,7 +49,7 @@ export const startRabbit = async (handler: HandlerMap) => {
   await channel.purgeQueue(receiveQueue);
   await channel.consume(
     receiveQueue,
-    async (e) => {
+    async (e: any) => {
```
Smrtnyk, line 52: Could we maybe type this better?

benawad:
this would break error catching for the ones returning a promise

PR #100, opened 2021-02-13
Fix button overlap on card

maskeynihal:
fix-card-blocked-issue


maskeynihal:
Solves #100

maskeynihal:
fixed #100

Issue #10, opened 2021-02-12
Remove yourself from "Speakers" without refreshing webpage


starbuck93:
![image](https://user-images.githubusercontent.com/5626027/107787852-a7b3f500-6d1d-11eb-8d3e-5df8a5f16f3a.png)

If you're a "Listener", click the button to become a "Speaker", but you can't go back to a "Listener" without refreshing the page.  Maybe the "Speaker" icon (above) could change to a "headphones" icon if you want to become a "Listener" again.

PR #102, opened 2021-02-13
let user to move to listener from speaker

Salman2301:
Not tested yet.
Just reuse the moderator **set_listener** logic.
( Hope the current user have the permission to move himself/herself to listener. )

Issue #10


benawad:
this already exists

Salman2301:
@benawad is it in process?
As a speaker i can't switch back to listener?
I have to refresh the page to go back to listener
Let me know if i am missing something 
Thank you

PR #107, opened 2021-02-13
solving the negative followers issue

NikolaTesla13:
I added an if statement when displaying the followers (if it's smaller than 0 display 0 else display the number of followers)

note: not tested but it should work, just an if


benawad:
I need to solve this on the Elixir side

PR #119, opened 2021-02-13
releases.exs: It's SENTRY_DSN, not SENTRY_DNS

AndreiJirohHaliliDev2006:
Also, I added an way for self-hosted instances to have their own environment variable for both Web UI and API server hosts. If left blank, it'll fall back to the defaults as https://dogehouse.tv and https://api.dogehouse.tv

### DIFF
--- a/kousa/config/releases.exs
+++ b/kousa/config/releases.exs
@@ -20,8 +20,10 @@ config :kousa,
       raise("""
       environment variable RABBITMQ_URL is missing.
       """),
-  web_url: "https://dogehouse.tv",
-  api_url: "https://api.dogehouse.tv",
+  web_url: 
+    System.get_env("WEB_URL") || "https://dogehouse.tv",
+  api_url: 
+    System.get_env("API_URL") || "https://api.dogehouse.tv",
   access_token_secret:
     System.get_env("ACCESS_TOKEN_SECRET") ||
       raise("""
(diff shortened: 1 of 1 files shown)


ianjaku:
Problem is this will break everyones local environment.

Perhaps we should phase it out gradually by adding a warning that SENTRY_DNS is depricated?

Ex:
``
if(System.get_env("SENTRY_DNS") != nil) do
    IO.warn("The SENTRY_DNS environment variable is deprecated, use SENTRY_DSN instead")
end

config :sentry,
  dsn:
    System.get_env("SENTRY_DSN") ||
    System.get_env("SENTRY_DNS") ||
      raise("""
      environment variable SENTRY_DSN is missing.
      """),
...
``

AndreiJirohHaliliDev2006:
> Problem is this will break everyones local environment.
> 
> Perhaps we should phase it out gradually by adding a warning that SENTRY_DNS is depricated?
> 
> Ex:
> 
> ```
> if(System.get_env("SENTRY_DNS") != nil) do
>     IO.warn("The SENTRY_DNS environment variable is deprecated, use SENTRY_DSN instead")
> end
> 
> config :sentry,
>   dsn:
>     System.get_env("SENTRY_DSN") ||
>     System.get_env("SENTRY_DNS") ||
>       raise("""
>       environment variable SENTRY_DSN is missing.
>       """),
> ```

@ianjaku Thanks for informing that matter. Next commit on the way.

PR #124, opened 2021-02-13
feat: toggle room visibility

### LINKED ISSUE #67: Feature Request : Toggle between public and private rooms
An option that enables rooms to toggle between private and public visibility even after they are created.

giuseppecampanelli:
Quick way to allow room privacy toggling which fixes #67 . Underlying code to switch between public & private can be cleaned up but I can only get to it later. If you need this out now, it has a first iteration.

### DIFF
--- a/kofta/src/vscode-webview/Routes.tsx
+++ b/kofta/src/vscode-webview/Routes.tsx
@@ -50,7 +50,11 @@ export const Routes: React.FC<RoutesProps> = () => {
         setCurrentRoom((cr) =>
           !cr || cr.id !== roomId ? cr : { ...cr, name, isPrivate }
         );
-        toast("room is now public", { type: "info" });
+        if (isPrivate) {
+          toast("room is now public", { type: "info" });
+        } else {
+          toast("room is now private", { type: "info" });
+        }
       },
       banned: () => {
         toast("you got banned", { type: "error" });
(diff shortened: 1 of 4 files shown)


kofta/src/vscode-webview/Routes.tsx
```diff
@@ -50,7 +50,11 @@ export const Routes: React.FC<RoutesProps> = () => {
         setCurrentRoom((cr) =>
           !cr || cr.id !== roomId ? cr : { ...cr, name, isPrivate }
         );
-        toast("room is now public", { type: "info" });
+        if (isPrivate) {
+          toast("room is now public", { type: "info" });
+        } else {
+          toast("room is now private", { type: "info" });
+        }
```
alii, line 57: You could convert this to a single line

```ts
toast(`Room is now ${isPrivate ? "private" : "public"}`, { type: "info" });
```

alii (changes requested):

benawad:
oops didn't mean to change status, but looks good to me, let me know when you want merge

kofta/src/vscode-webview/Routes.tsx
giuseppecampanelli: Changed!

giuseppecampanelli:
@benawad ok to merge

PR #144, opened 2021-02-13
add background color to drop-down to match dark mode UI

leonyanagida:
because Ben loves dark mode


leonyanagida:
Not sure if I should delete the package-lock.json here

benawad:
does the color of the option text need to change? Also you should keep the package-lock.json as is

leonyanagida:
> does the color of the option text need to change? Also you should keep the package-lock.json as is

No, the option text doesn't need to be changed. Unless you specifically want the text to be a different color.
Thanks, I deleted the package-lock.json file.

benawad:
Yeah, can you put it back

leonyanagida:
> Yeah, can you put it back

Whoops, it's back now.

benawad:
reset it, so there are no package-lock changes in this pr

Issue #116, opened 2021-02-13
Push to talk Button


ya5hpat3l:

MrMalloc:
As we have noticed this would be very helpful when someone without a headset is in the room. Right now such members have to remember to mute themselves after talking so there is no echo over their speakers.

nadirabbas:
@benawad I can work on this.

PR #164, opened 2021-02-14
feat: push-to-talk

nadirabbas:
Solves #116

### DIFF
--- a/kofta/src/vscode-webview/components/KeybindListener.tsx
+++ b/kofta/src/vscode-webview/components/KeybindListener.tsx
@@ -11,6 +11,7 @@ export const KeybindListener: React.FC<KeybindListenerProps> = ({}) => {
 
   return (
     <GlobalHotKeys
+      allowChanges={true}
       keyMap={keyMap}
       handlers={useMemo(
         () => ({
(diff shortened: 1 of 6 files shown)


kofta/src/vscode-webview/components/KeybindListener.tsx
```diff
@@ -23,6 +23,17 @@ export const KeybindListener: React.FC<KeybindListenerProps> = ({}) => {
...
+            if (!e) return
+            
+            const { muted, set } = useMuteStore.getState();
+            const mute = e.type === 'keyup' 
+            wsend({
+              op: "mute",
+              d: { value: mute },
+            });
+            set({ muted: mute });
+          }
```
benawad, line 36: don't you need 2 handlers? 1 for keyup and 1 for keydown?

benawad (changes requested):

kofta/src/vscode-webview/components/KeybindListener.tsx
nadirabbas: @benawad no it can be done in the same handler, we set the mute states based on if the key was pressed down or released.

nadirabbas: qEnyfKRnYB

benawad:
Oh I see, that fires on both events

PR #188, opened 2021-02-14
ReadME badge init

LeulAria:
used https://shields.io/ badge generator, can add other badges on demand...


benawad:
The YouTube like one is kind of interesting, can you have that link to the video too. The other ones are clutter IMO and I'd rather them not be there.

OfshoX:
you have the like twice

PR #203, opened 2021-02-14
Feature: unread messages number

carlosqsilva:
Captura de Tela 2021-02-14 às 17 45 32


benawad:
I'm undecided on whether I want to show this. I think it's a little cleaner just to show only the bubble, but I will think on this.

tanay-pingalkar:
> I'm undecided on whether I want to show this. I think it's a little cleaner just to show only the bubble, but I will think on this.

you should add this feature

SamiTriki:
That could be a setting, I believe slack has it set up this way

benawad:
You can't really see that message icon well anymore. I'm indifferent about showing the number, but if you want to go with a different design option, you can reopen.

PR #219, opened 2021-02-15
Added background colour to option

### LINKED ISSUE #166: Mic list dropdown: options contrast isn't enough
image

RazerMoon:
Resolves #166

Before:
before

After:
after

### DIFF
--- a/kofta/src/css/vscode.css
+++ b/kofta/src/css/vscode.css
@@ -131,7 +131,7 @@ input {
 }
 
 option {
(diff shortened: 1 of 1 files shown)


leonyanagida:
How about trying the color `var(--vscode-dropdown-background);`  instead? 
The dropdown color is in the CSS name itself.

RazerMoon:
@leonyanagida fixed it, didn't notice that value 😂

PR #232, opened 2021-02-15
feat: mention, anchor in chat

nadirabbas:
- Mention other users by displayName or username
- Clicking a mentioned username opens up user's profile modal
- if somebody mentions you, you see an `@` sign on the bottom navigation (not done yet)
- Clickable links in chat

It may be appropriate to merge this pull request when ben does his pattern matching magic in Kousa for token structure. (I have temp removed it for testing)

works something like this:

i6Jf2S5IqG


benawad:
This looks good, I will integrate this when I do the backend validation.

mohammedgad:
Hi @benawad I will work on the BE validation.

benawad:
@mohammedgad ok, the main thing to validate is messages should be less than 512 characters, the type should be one of the known ones, and the tokens should not include in unknown fields.

mohammedgad:
@benawad what happens if a token is invalid or a message exceeds 512 char, should we discard it?

benawad:
yeah for now just ignore it

kofta/src/vscode-webview/modules/room-chat/RoomChatList.tsx
```diff
@@ -1,48 +1,97 @@
...
+                {m.displayName}
+              </button>
+              <span className={tw`mr-1`}>: </span>
+            </>
+          ) : null}
           {m.tokens.map(({ t, v }, i) =>
             t === "text" ? (
-              <span className={tw`flex-1`} key={i}>
-                {v}
+              <span className={tw`flex-1 m-0`} key={i}>
```
benawad, line 60: can you refactor this into a switch statement so its cleaner to read

kofta/src/vscode-webview/pages/Home.tsx
```diff
@@ -70,19 +72,28 @@ export const Home: React.FC<HomeProps> = () => {
...
         ) : null}
-        {rooms.map((r) =>
+        {rooms.map(r =>
           r.id === currentRoom?.id ? null : (
             <div className={tw(`mt-4`)} key={r.id}>
               <RoomCard
                 onClick={() => {
                   wsend({ op: "join_room", d: { roomId: r.id } });
+                  // Show joined message in chat
+                  setTimeout(() => {
```
benawad, line 82: don't send this from the client-side, I'll trigger it from the backend

nadirabbas: Yeah O kay I will remove it now

benawad:
@nadirabbas I tested this out locally, nice work!

1. The mention popup doesn't show if I try to mention a second user in the same message
2. Can you split the mention logic into it's own zustand store

nadirabbas:
Hey Ben, sure! will split it into separate files. 
I am not sure about the first point you mention because it used to work, maybe messed up while changing other things. Will also fix it.

nadirabbas:
@benawad actually multiple mentions in a single message are working on the live server.

benawad:
My bad, I think I just needed more people in the room

PR #252, opened 2021-02-16
Chat href

ridarf:
Wasn't able to run server on my own.

Was inconvenient to manually visit shared link in social spaces, added Regex matching to URLs, mapping them to anchor tags.


nadirabbas:
#232 this is already done, sorry for not informing, we kinda overlapped

ridarf:
@nadirabbas Got it, didn't realize and there were a few issues with my code. Glad to know that's been implemented.

PR #277, opened 2021-02-18
add BE link, mention, and message length validation

mohammedgad:
BE validation for https://github.com/benawad/dogehouse/pull/232 

- Validates links using elixir URI.
- Validates message length to be less than or equal to 512.

### DIFF
--- a/kousa/lib/business-logic/room_chat_logic.ex
+++ b/kousa/lib/business-logic/room_chat_logic.ex
@@ -57,19 +57,10 @@ defmodule Kousa.BL.RoomChat do
   end
 
   @doc """
-    validate_token/1 validates each token by type and return {:ok, token} if the token is valid
+    validate_token/1 validates a token by type and return {:ok, token} if the token is valid
     otherwise return :invalid
   """
-  defp validate_token(token = %{"t" => "text", "v" => _}), do: {:ok, token}
-  defp validate_token(token = %{"t" => "mention", "v" => mention}) do 
-    mention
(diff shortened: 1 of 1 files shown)


nadirabbas:
@mohammedgad There is also a token type called "event", can you please add that too before this is merged?

kousa/lib/business-logic/room_chat_logic.ex
```diff
@@ -1,33 +1,97 @@
...
+  end
+
+  defp validate_token(_), do: :invalid
+
+  defp valid_url?(%URI{host: host, scheme: scheme}) when is_binary(host) and is_binary(scheme), do: true
+  defp valid_url?(_), do: false
+
+  defp valid_mention?(mention) do
+    mention
+    |> Data.User.get_by_username()
```
benawad, line 90: I want to avoid a db lookup on message sends, I'm ok if someone does an @ for a non-existent user, so you can remove this

benawad:
don't add the event type, that should only be triggered by the backend not the frontend

nadirabbas:
Okay I will remove it

kousa/lib/business-logic/room_chat_logic.ex
nadirabbas: it won't work for a non-existing user anyway.

mohammedgad: @benawad @nadirabbas yea that sounds reasonable according to business. I will remove it.

PR #281, opened 2021-02-18
feat: search rooms

### LINKED ISSUE #279: Add functionality on homepage to search chat rooms

nadirabbas:
Fixes #279

image


benawad:
I don't think we need a search until there are more rooms than can be loaded on the front page in which case we will have to do a db query

nadirabbas:
Got it

Issue #278, opened 2021-02-18
Use smaller Github avatars


nadirabbas:
GitHub allows adding an `s` parameter in the avatar URL. We can set it to different values at different places in the App. 
Such as, on the room page, the pictures are smaller, we can use something like `s=50`.  And if a user opens the profile modal, then we can use a bigger number.

nadirabbas:
#282

PR #282, opened 2021-02-18
Use smaller Github avatars

nadirabbas:
#278


benawad:
I think its better to load 1 img, let the browser cache it for all the places.

Issue #298, opened 2021-02-18
DogeHouse is no longer loading on iOS Safari


younesaassila:
When accessing [DogeHouse.tv](https://dogehouse.tv) from Safari on iOS, a blank dark-gray page is shown.

This issue started appearing today.

Here's a screenshot: ![Screenshot](https://i.imgur.com/JxCdDqW.png)

**OS:** iOS 14.4

nadirabbas:
@benawad I think it is in the regex for mention matching. iOS has been known for not supporting some regex patterns. Sadly I don't have Apple products so I couldn't test.

overlisted:
It might be because something on the way to final react-dom `render()` call is broken

ajmnz:
It definitely has to do with some regex error. 

```
SyntaxError: Invalid regular expression: invalid group specifier name
```

Working on it.

ajmnz:
Got it. Error is coming from the package `normalize-url`. `kofta/node_modules/normalize_url/index.js` @ line 125.

Commit cb2f0d1b276819189ccb4e7fc30864ebc3a379f2 introduced the bug.

Will downgrade it to normalize-url@4 and test asap if that solves the issue.

benawad:
commented out normalize-url for now

ajmnz:
> commented out normalize-url for now

Made a PR

PR #300, opened 2021-02-18
Fix: Downgrade to normalize-url@4 to prevent regex bug on Safari

ajmnz:
Solves #298. DogeHouse was breaking on Safari (macOS & iOS) due to regex lookahead and lookbehind not being supported by the browser.


ajmnz:
Can confirm that DogeHouse is up again on Safari (macOS & iOS) 👍🏼

Issue #306, opened 2021-02-19
Notification Sound when you get a message 


Rishabh-malhotraa:
## WHY?
I was alone in a room and had the doghouse link open in the background with my mic muted, people came into my room and wrote in the chat but I was not aware, it would be great if you get a voice sound every time some user sends a message (also have the ability to turn off this feature )

![image](https://user-images.githubusercontent.com/54576074/108469380-1f4dcb00-72ae-11eb-987a-dafec70d440d.png)

mohammedgad:
We need that! @nadirabbas what do you think?

nadirabbas:
Yeah I think, it should be toggleable tho

rh3xp:
It should notify especially when someone mentions you, like a fuzzy search of all the usernames in the current chat window.

nadirabbas:
PR on the way

nadirabbas:
#308

PR #308, opened 2021-02-19
Feature/notification sound

nadirabbas:
Play a notification sound when the chat sidebar is closedor the user is inactive for 45 seconds (AFK) and somebody mentions your name. This setting is toggleable and stored in the local storage.

Increase your speaker's volume to watch this video:

https://user-images.githubusercontent.com/38838675/108507199-c5fc9080-72db-11eb-8821-401dc7f20523.mp4


Rishabh-malhotraa:
closes #306

benawad:
I think I want a simpler condition for when the sound occurs: if chat is closed or document is not focused (https://stackoverflow.com/questions/17389280/check-if-window-has-focus/17389334)

nadirabbas:
@benawad  Sometimes the document has the focus but the user is away from the keyboard for some time (45 seconds in this case). I specifically did it for the AFK problem. IF we don't need AFK sound, then I can remove the AFK timer completely.

benawad:
I don't want an AFK timer atm

nadirabbas:
@benawad Please review the changes when you get time.

kofta/src/vscode-webview/Routes.tsx
```diff
@@ -46,6 +50,30 @@ export const Routes: React.FC<RoutesProps> = () => {
...
+              t.t === "mention" &&
+              t.v?.toLowerCase() === me?.username?.toLowerCase()
+          ).length
+        ) {
+          useRoomChatStore.getState().setIAmMentioned(iAmMentioned + 1);
+        }
+      },
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [me, iAmMentioned, open]);
```
benawad: instead of creating a new useEffect, do `const {open, iAmMentioned} = useRoomChatStore.getState()` to get the most recent values

kofta/src/vscode-webview/components/BottomVoiceControl.tsx
```diff
@@ -40,7 +40,11 @@ export const BottomVoiceControl: React.FC<BottomVoiceControlProps> = ({
   const { muted, set } = useMuteStore();
   const [{ canSpeak, isCreator }] = useAtom(myCurrentRoomInfoAtom);
   const [settingsOpen, setSettingsOpen] = useState(false);
-  const [toggleOpen, newUnreadMessages, iAmMentioned] = useRoomChatStore(s => [
+  const [
+    toggleOpen,
+    newUnreadMessages,
+    iAmMentioned,
+  ] = useRoomChatStore((s) => [
```
benawad: also before we go to far with chat logic, can you split mention state into its own store

kofta/src/vscode-webview/components/VoiceSettings.tsx
```diff
@@ -7,9 +7,11 @@ import { Button } from "./Button";
 import { MuteKeybind, PTTKeybind } from "./keyboard-shortcuts";
 import { VolumeSlider } from "./VolumeSlider";
 
-interface VoiceSettingsProps {}
+interface VoiceSettingsProps {
+  className?: string;
+}
 
-export const VoiceSettings: React.FC<VoiceSettingsProps> = ({}) => {
+export const VoiceSettings: React.FC<VoiceSettingsProps> = ({ className }) => {
```
benawad: instead of passing a className to VoiceSettings, just wrap the component in a `<div />` with the `className` you want to apply

kofta/src/vscode-webview/pages/MyProfilePage.tsx
```diff
@@ -36,8 +37,8 @@ export const MyProfilePage: React.FC<MyProfilePageProps> = ({}) => {
         </div>
       </Backbar>
       {me ? <UserProfile profile={me} /> : <div>probably loading...</div>}
-      <h1 className={tw`py-8 text-4xl`}>Voice Settings</h1>
       <VoiceSettings />
+      <ChatSettings className={tw`mt-8`} />
```
benawad: same thing, wrap with a div instead of passing className

kofta/src/vscode-webview/types.ts
```diff
@@ -25,6 +25,7 @@ export type User = {
   currentRoom?: Room;
   followsYou?: boolean;
   youAreFollowing?: boolean;
+  setshouldPlayChatNotificationSound: boolean;
```
benawad: `setshouldPlayChatNotificationSound` -> `setShouldPlayChatNotificationSound`

nadirabbas: I didn't get this one, what do you mean?

kofta/src/vscode-webview/Routes.tsx
nadirabbas: Great thanks.

nadirabbas: @benawad I have to `useEffect` for the `me` atom. Because when the `addMultipleWsListener` registers, `me` is null

kofta/src/vscode-webview/components/BottomVoiceControl.tsx
nadirabbas: Sure

kofta/src/vscode-webview/Routes.tsx
nadirabbas: Should I remove this logic from Routes.tsx and put it somewhere, that way we can get the latest value of `me` I think.

nadirabbas:
@benawad I have to use `useEffect` for the `me` atom. Because when `addMultipleWsListener` registers, `me` is null. Or is there another way?

I could actually do this logic in another file where I don't have a function registering like `addMultipleWsListener` but a function that actually gets new values.

kofta/src/vscode-webview/Routes.tsx
benawad: I see, lets do a quick fix by sticking current me value in a ref
```
const meRef= useRef(me);
meRef.current = me
```

nadirabbas: @benawad Okay it works that way 🎉

nadirabbas:
@benawad All changes are done, and it works now.
You first have to enable this option from settings.

kofta/src/vscode-webview/Routes.tsx
```diff
@@ -58,6 +71,16 @@ export const Routes: React.FC<RoutesProps> = () => {
       },
       new_chat_msg: ({ msg }) => {
         useRoomChatStore.getState().addMessage(msg);
+        if (
+          (!open || !document.hasFocus()) &&
```
benawad, line 75: this is using a stale value for open, don't you need to grab it from getState()?

nadirabbas:
@benawad I have fixes merge conflicts

kofta/src/vscode-webview/Routes.tsx
nadirabbas: @benawad It is using `getState`

nadirabbas: Oh you mean use it directly from state inside the `new_chat_msg`?

nadirabbas: Yeah my bad, couldn't pick it up because the document was always out of focus as I was testing. Fixed

PR #341, opened 2021-02-20
Fix padding

RazerMoon:
The wrapper element currently has padding, which creates an ugly gap. I've now made that a default style, and overwritten it in most pages to improve the look.

Before             |  After
:-------------------------:|:-------------------------:
  |  

## Another:

Before             |  After
:-------------------------:|:-------------------------:
  |


DGKSK8LIFE (approved):
looks a lot better

kofta/src/vscode-webview/pages/BanUsersPage.tsx
```diff
@@ -24,37 +24,39 @@ export const BanUsersPage: React.FC<SearchUsersProps> = ({}) => {
   }
 
   return (
-    <Wrapper>
+    <Wrapper style={{ marginBottom: "auto" }}>
```
benawad: instead of putting this on every wrapper, should we just remove the padding from the wrapper component?

DGKSK8LIFE: Agreed.

RazerMoon: I think that would be a good idea. I get why the padding is there, but the backbar is inside the wrapper on every page, and applying padding to a backbar doesn't look good. That method only really works on the home and login page, because there is no backbar there.

benawad: Ok then remove the padding from `<Wrapper />` and create a  `<BodyWrapper />` with the padding and use that on the main content for all the pages

RazerMoon:
Done, hope it's alright that I changed the width on the modal by the way. Now it looks better on mobile, barely noticeable change on desktop.

| Before | After           |
| ------------- |:-------------:|
|       |  |

benawad:
So does `RoomPage` and `Login` not need side padding on mobile?

RazerMoon:
That's my bad 😅, should be good now

PR #343, opened 2021-02-21
Better footer and buttons

RazerMoon:
## Footer

Slightly improved look on mobile

| Before | After |
|:----:|:----|
|    before   |     after  |

## Buttons

I think the buttons look too big on desktop. Now, there is a new "slim" variant, which is also responsive.

Also, made the font size on the "Create Room" button slightly larger.

| Responsive showcase | Slim |
|:----:|:----|
|   awdaw   |  slim  |


nadirabbas:
@RazerMoon can't we remove a couple of footer links on mobile, @benawad ?

RazerMoon:
On a side note, how about changing the front page design to something like this?

| Before   | After  |
|:---:|:---:|
| before | after |

Ofc the new design would go back to the old one on smaller screens.

benawad:
@RazerMoon I don't really like that design change, unless some other stuff changed on the landing page

PR #345, opened 2021-02-21
Feature/linkify username

nadirabbas:
Putting a URL in this field makes usernames clickable.

image
image


kousa/priv/repo/migrations/20210221014438_external_profile_link.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Beef.Repo.Migrations.ExternalProfileLink do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :externalProfileLink, :text, default: nil, allow_nil: true
```
mohammedgad: that should be a string.

kousa/lib/beef/user.ex
```diff
@@ -48,18 +50,19 @@ defmodule Beef.User do
...
 
   def edit_changeset(user, attrs) do
     user
-    |> cast(attrs, [:id, :username, :bio, :displayName])
+    |> cast(attrs, [:id, :username, :bio, :displayName, :externalProfileLink])
     |> validate_required([:username, :bio, :displayName])
     |> validate_length(:bio, min: 2, max: 160)
     |> validate_length(:displayName, min: 2, max: 50)
     |> validate_format(:username, ~r/^(\w){4,15}$/)
+    |> validate_format(:externalProfileLink, ~r/^(http|https):\/\/[^ "]{1,255}$/)
```
mohammedgad: hi @nadirabbas I think the URL edit can't be nil because the validation `{1,255}` can you try to make it `{0,255}`"?

mohammedgad (changes requested):

kousa/lib/beef/user.ex
nadirabbas: oh damn

nadirabbas: It works @mohammedgad. Well, that was a dumb mistake. Nice catch 🎉

benawad:
I don't want a separate link for username atm. I only want to linkify bio

nadirabbas:
@benawad I guess I misunderstood yesterday :/

nadirabbas:
I Will do that instead, sorry I misunderstood

benawad:
it's all good, I could have been clearer

PR #361, opened 2021-02-22
Bugfix/sound default

nadirabbas:
Make notification sound play by default (mentions in chat)


DGKSK8LIFE (approved):
Better looking ternaries.

nadirabbas:
> Better looking ternaries.

yeah 😅

PR #377, opened 2021-02-23
I think outline is not looking good

yakohere:


benawad:
I'm pretty sure it's needed for accessibility?

yakohere:
> I'm pretty sure it's needed for accessibility?

you seem to be right.

PR #381, opened 2021-02-23
Kofta: Fully switch to Tailwind

overlisted:
- Switched to TailwindCSS instead of the twind package
- Managed to get rid of almost every single `style` attribute (just 15 left over)
- Removed the "css" directory completely
- Installed CRACO

My next PR is probably gonna be a switch to yarn and Parcel (or webpack).


overlisted:
Regarding the conflicts: _Please_ resolve them in favor of Tailwind and keep in mind that `tw` template tag is no longer required.

nadirabbas:
@benawad styling of npm packages is no longer working.

PR #382, opened 2021-02-23
Linkify bio

nadirabbas:


kofta/src/vscode-webview/components/UserProfile.tsx
```diff
@@ -115,7 +117,23 @@ export const UserProfile: React.FC<UserProfileProps> = ({
...
-      <div className={`mb-4`}>{profile.bio}</div>
+      <div className="mb-4">
+        {profile.bio?.split(" ").map((chunk, i) => {
+          return linkRegex.test(chunk) ? (
+            <a
+              key={i}
+              href={normalizeUrl(chunk)}
+              target="_blank"
+              rel="noreferrer"
+              className="text-tmpC4 p-0"
```
benawad: I think `tmpC4` is gone

DGKSK8LIFE: Yeah might want to double-check the css

nadirabbas: when i wrote it the new breaking PR wasn't merged. so didn't know, fixing.

PR #389, opened 2021-02-23
Fix RoomCard styles

overlisted:
in my offense i couldn't test the component without putting it in `Home.tsx` with a dummy room data


DGKSK8LIFE (commented):
It's probably best to remove the dummy data

overlisted:
It's only shown in debug mode

PR #391, opened 2021-02-23
Show a toast when people leave or join

overlisted:
Implements #6


benawad:
I don't want this

overlisted:
lol it was in your projects

benawad:
if its in priority 4, I don't want it now, and will consider adding at a later point

PR #398, opened 2021-02-24
Following list: show offline users

### LINKED ISSUE #256: Active users being followed
If none of the users I follow are online, the list just says "no users found". It would be better if there are two sections called "Offline Users" and "Online Users" with the offline ones being faded and below the "Online Users" section.

nadirabbas:
Fixes #256

- Show green dot if the user is online
- Sort following/followers by online status

image


benawad:
instead of having 2 tabs, I think I would prefer a green dot or something to indicate they are online

nadirabbas:
@benawad Got it

nadirabbas:
@benawad conflicts resolved

benawad:
Sorting on the frontend is not enough because we are using pagination, will need to sort in the db query

I didn't realize this replaces the people icon, the point of that is to see what room the people you follow are in.

`FollowingOnlineList.tsx` needs to stay, but the backend query needs to grab non-online users too

nadirabbas:
@benawad will do that

PR #399, opened 2021-02-24
Minor idioms demo

ityonemo:


kousa/lib/data-layer/follower_data.ex
```diff
@@ -136,10 +136,14 @@ defmodule Kousa.Data.Follower do
...
   def insert(data) do
-    case %Beef.Follow{}
-         |> Beef.Follow.insert_changeset(data)
-         |> Beef.Repo.insert() do
+    %Beef.Follow{}
+    |> Beef.Follow.insert_changeset(data)
+    |> Beef.Repo.insert()
+    |> case do
       {:ok, _} ->
+        # TODO: eliminate N+1 by setting up changesets
```
benawad, line 144: what do you mean by this?

ityonemo: You can set up associations in the schema so that ecto fashions the join and correctly populates the value in one trip to the database.

benawad: won't it always just be 1 insert to follower table and 1 update to user table?

ityonemo: 2 database actions, but 1 transaction.

PR #401, opened 2021-02-24
Starts tests on database

ityonemo:
note that some of these tests are failing (for demonstration purposes) and you may want to consider activating gh actions to block PRs into staging with failing tests.


ityonemo:
also note that over time you are going to want to change away from camelCase on those schemas to snake_case.  Everything seems to be working, but elixir really expects you to be using snake_case on things, so this could cause problems down the line.

kousa/mix.exs
```diff
@@ -21,16 +22,18 @@ defmodule Kousa.MixProject do
     ]
   end
 
-  # Run "mix help deps" to learn about dependencies.
   defp deps do
     [
       {:amqp, "~> 1.0"},
+      # TODO: consider switching to Registry
```
benawad, line 28: doesn't this use Registry underneath the hood? What would the advantage of using Registry directly?

ityonemo: it's just a bias, to not use libraries that are such a thin shim over stdlib features.

1 - blocks learning about the stdlib
2 - adds a barrier for contributors to immediately grok
3 - upgrade pathways to more advanced things (you will soon ;-) have to pick swarm or horde - i recommend horde) and you will have to decouple first and then upgrade.

ityonemo: anyways it's not super important right now, plus you want tests on everything before you refactor =P

PR #412, opened 2021-02-25
guards result, logs inbound settings, adds example typespec

ityonemo:


kousa/lib/routes/github_auth.ex
```diff
@@ -38,14 +39,16 @@ defmodule Kousa.GitHubAuth do
...
     with true <- Kousa.Caster.bool(Application.get_env(:kousa, :is_staging)),
          state <- Map.get(conn.query_params, "state", ""),
          {:ok, json} <- Base.decode64(state),
-         {:ok, %{"redirect_base_url" => redirect_base_url}} <- Poison.decode(json) do
+         {:ok, %{"redirect_base_url" => redirect_base_url}} when is_binary(redirect_base_url) <- Poison.decode(json) do
       redirect_base_url
     else
-      _ ->
+      other ->
+        Logger.warn("unknown inbound base_url setting #{inspect other}")
```
benawad, line 51: you don't need to log this, the `redirect_base_url` is only used in staging environment. In prod this always runs

Issue #413, opened 2021-02-25
Wrong color when somebody mentions you


nadirabbas:
When somebody mentions you, the color used to be bg-blue. Now it is gray-ish.
![image](https://user-images.githubusercontent.com/38838675/109106510-b785fd00-7751-11eb-9276-ec33e541f38e.png)

nadirabbas:
#416

PR #416, opened 2021-02-25
Fix chat styling

### LINKED ISSUE #414: Woahhhh
image

nadirabbas:
Fixes #414 and #413 

(I had to pass in `className` prop to `Avatar`, otherwise, I had to use `style` attr, which we're afraid of. In this case, classes on the wrapper won't solve the issue)

### DIFF
--- a/kofta/src/vscode-webview/components/Avatar.tsx
+++ b/kofta/src/vscode-webview/components/Avatar.tsx
@@ -6,6 +6,7 @@ interface AvatarProps {
   active?: boolean;
   circle?: boolean;
   usernameForErrorImg?: string;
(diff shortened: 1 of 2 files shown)


tanay-pingalkar:
have you made the scroll bar like it was before

nadirabbas:
@tanay-pingalkar yes sire

tanay-pingalkar:
And can you make height of emoji feature half of what it is now.....bcoz its looking ugly, i had made its height small but some changed that..make overflowY hidden in tailwind if possible..

tanay-pingalkar:
please

PR #422, opened 2021-02-25
Fix long room titles on the home page

### LINKED ISSUE #421: Horizontal Scrolling on Long Group Title
When I'm wringing a long group name it breaks the UI.
bug_dogehouse

nadirabbas:
Fixes #421 

| Before | After |
| ------------- | ------------- |
| image | image|

### DIFF
--- a/kofta/src/vscode-webview/components/RoomCard.tsx
+++ b/kofta/src/vscode-webview/components/RoomCard.tsx
@@ -49,7 +49,9 @@ export const RoomCard: React.FC<RoomProps> = ({
     <div>
       <button
         onClick={onClick}
-        className={`w-full ${active ? "bg-simple-gray-4d" : "bg-simple-gray-33"} hover:bg-simple-gray-69 active:bg-simple-gray-23 py-2.5 px-5 rounded-lg`}
+        className={`w-full ${
+          active ? "bg-simple-gray-4d" : "bg-simple-gray-33"
+        } hover:bg-simple-gray-69 active:bg-simple-gray-23 py-2.5 px-5 rounded-lg`}
       >
         <div className={`flex text-white`}>
           <div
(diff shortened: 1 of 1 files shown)


kofta/src/vscode-webview/components/RoomCard.tsx
```diff
@@ -49,7 +49,9 @@ export const RoomCard: React.FC<RoomProps> = ({
     <div>
       <button
         onClick={onClick}
-        className={`w-full ${active ? "bg-simple-gray-4d" : "bg-simple-gray-33"} hover:bg-simple-gray-69 active:bg-simple-gray-23 py-2.5 px-5 rounded-lg`}
+        className={`w-full ${
+          active ? "bg-simple-gray-4d" : "bg-simple-gray-33"
```
jobveldhuis, line 53: This code reformatting does not change functionality. I think I'd prefer this PR without this change.

kofta/src/vscode-webview/components/RoomCard.tsx
```diff
@@ -58,7 +60,7 @@ export const RoomCard: React.FC<RoomProps> = ({
               WebkitBoxOrient: "vertical",
               WebkitLineClamp: 3,
             }}
-            className={`text-left flex-1 text-xl text-simple-gray-d9 text-ellipsis overflow-hidden break-words`}
+            className={`text-left flex-1 text-xl text-simple-gray-d9 text-ellipsis overflow-hidden break-all`}
```
jobveldhuis, line 63: This seems to be a good and efficient fix. LGTM.

jobveldhuis (changes requested):
As stated, I prefer this PR without the reformatting of the code that does not change the behaviour.

kofta/src/vscode-webview/components/RoomCard.tsx
nadirabbas: this is just prettier doing it, so doesn't really matter either way.

Issue #436, opened 2021-02-26
react frontend won't run locally


tejxv:
well, i tried to run the script ```npm start:staging``` but it throws an error so i ran ```npm run start:staging``` and it partially worked but the error this time was
```
PS C:\Users\GTX\ben\fork\dogehouse\kofta> npm run start:staging

> kofta@0.1.0 start:staging C:\Users\GTX\ben\fork\dogehouse\kofta
> REACT_APP_IS_STAGING=true REACT_APP_API_BASE_URL=https://doge-staging.stripcode.dev craco start

'REACT_APP_IS_STAGING' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! kofta@0.1.0 start:staging: `REACT_APP_IS_STAGING=true REACT_APP_API_BASE_URL=https://doge-staging.stripcode.dev craco start`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the kofta@0.1.0 start:staging script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
 ```
idk what's wrong, and im on windows btw

benawad:
npm i -D cross-env

Then put cross-env in front of the staging command.

tejxv:
that worked! thank you!

benawad:
can you pr it for other window users

tejxv:
okay sure

PR #437, opened 2021-02-26
add cross-env

tejxv:
#436


benawad:
cross-env works on Mac and Linux so you can just stick it in there and it'll work for everyone. You don't need to update docs, my bad for being unclear

tejxv:
lol i was actually gonna ask about that but anyways no problem i'll do that tomorrow because it's 3 in the morning and im off to bed

PR #442, opened 2021-02-27
A Very Serious PR adding the Doge to a button

TheTimemaster:
A sliding one that appears in random localization when you hover over the button, I must add.

I think this will have a very positive impact on user retention in the near future.

Zoooooop.


benawad:
Can you make it a prop you pass in to enable this behavior, I don't won't this on every button, but I want this on some of them

TheTimemaster:
Added dogeProbability prop that gives even a bit more flexibility.

PR #447, opened 2021-02-27
ts-node support for shawarma

shmuelhizmi:
ts-node is just nicer to work with, stuff like getting your errors traced to the right position and not having to run  watch + dev every time


nadirabbas:
@shmuelhizmi great. Can you please reset `package-lock.json`

benawad:
@nadirabbas if you install deps its ok if package-lock changes

nadirabbas:
my bad 😅

PR #452, opened 2021-02-27
kofta - desktop UI changes: move the main view and chat view into css…

shmuelhizmi:
desktop UI changes
before:
image
after:
image

The mobile view looks the same.

Note:
I know the app is attended to look flat (which make senic because the app was originally built for vscode which use a flat design)
and using rounded cards with shadows kinda break the design, therefore I didn't change that app inner view (there is actually no change to how to mobile app looks) - I just containerized the desktop view with a container that supposes to look like a smartphone (rounded corners and a border).


benawad:
login is off center now https://deploy-preview-452--admiring-bose-46a41d.netlify.app/

shmuelhizmi:
@benawad yep forget to check, it is fixed now.

PR #454, opened 2021-02-27
prevent user from sending more than 1 message a second in chat on the frontend

### LINKED ISSUE #152: prevent user from sending more than 1 message a second in chat on the frontend (this is already on backend)

damian-kociszewski:
Fixes #152

### DIFF
--- a/kofta/src/vscode-webview/modules/room-chat/RoomChatInput.tsx
+++ b/kofta/src/vscode-webview/modules/room-chat/RoomChatInput.tsx
@@ -1,5 +1,6 @@
 import { useAtom } from "jotai";
-import React, { createRef, useState } from "react";
+import React, { useEffect, useRef, useState } from 'react';
+import { toast } from 'react-toastify';
 import { wsend } from "../../../createWebsocket";
 import { meAtom } from "../../atoms";
 import { modalAlert } from "../../components/AlertModal";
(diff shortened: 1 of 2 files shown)


kofta/src/vscode-webview/modules/room-chat/RoomChatInput.tsx
```diff
@@ -63,45 +59,63 @@ export const RoomChatInput: React.FC<ChatInputProps> = ({}) => {
...
+          modalAlert("You got banned from chat");
           return;
-        }
-        if (me.id in useRoomChatStore.getState().bannedUserIdMap) {
-          modalAlert("you got banned from chat");
+      }
+      
+      if (messageTimeout) {
+          if (!toast.isActive("message-timeout")) {
+              toast(
```
damian-kociszewski, line 84: let the user know why the message has not been sent

benawad:
instead of timeout based, can you set the date of the last message sent and check against that

nadirabbas:
@benawad that way we can probably also add timestamps in the chat. maybe in a tooltip.

benawad:
that'll be a separate thing, timestamped on the server

damian-kociszewski:
> instead of timeout based, can you set the date of the last message sent and check against that

sure

PR #456, opened 2021-02-27
improved text truncating (room page -> room name & usernames);

damian-kociszewski:
I feel like that looks better


kofta/src/vscode-webview/components/UserNode.tsx
```diff
@@ -66,10 +66,10 @@ export const UserNode: React.FC<UserNodeProps> = ({
           </div>
         ) : null}
       </div>
-      <div className={`mt-2 flex items-center`}>
-        <div className={`text-sm`}>
-          {(u.displayName || "").split(" ")[0]?.slice(0, 9)}
-        </div>
+      <div className={`mt-2 flex w-full justify-center items-center truncate`}>
+        <span className={`text-sm truncate`}>
+          {(u.displayName || "Anonymous").trim().split(" ")[0]}
```
benawad: can you make this their `username`, `Anonymous` is going to be used for something else later

PR #471, opened 2021-02-28
Add prettier pre-commit hook

### LINKED ISSUE #460: add a prettier pre-commit hook

nadirabbas:
Fixes #460

The reason why this has so many changes is `pre-commit` fixed some files.


WikiRik:
Any reason why so many rules are turned off?

nadirabbas:
this is just preference, they can be turned on. This is the AirBNB standard.

WikiRik:
I thought that extending airbnb was enough to use their rules. So it is unnecessary to turn off the rules. But maybe I am wrong?

.eslintrc.js
```diff
@@ -0,0 +1,287 @@
+module.exports = {
+  env: {
+    browser: true,
+    es2021: true,
+  },
+  extends: ["plugin:react/recommended", "airbnb"],
```
WikiRik, line 6: I see that prettier is part of the dependencies, but it's not extended nor put in the plugins. I think you can copy the recommended configuration of the prettier plugin; https://github.com/prettier/eslint-plugin-prettier#recommended-configuration

nadirabbas: This is actually the first time I am doing this manually, so I was bound to fail lolll

nadirabbas: will fix when I get to it.

WikiRik: No worries, nice that you want to work on this! Maybe you can check out the prettier documentation as well, it has a page on pre-commit hooks; https://prettier.io/docs/en/precommit.html

WikiRik:
@benawad can I ask why you merged this? it seemed that @nadirabbas was still working on this

nadirabbas:
@WikiRik It is still working tho. I copied these rules from a site. where t hey were recommended.

nadirabbas:
@WikiRik you can make another PR probably to optimize the rules.

PR #481, opened 2021-02-28
feat: whispers in chat

### LINKED ISSUE #402: send private messages in the room chat
Sometimes you want to say something to someone in secret, but you can't because everything in chat is visible to everyone.

I suggest something like @user #whisper {insert message here} - only the users you've tagged should be able to see it.

This is quicker and easier to implement than an entire private chat thing - so if that's in the works this can serve as a placeholder of sorts.

nadirabbas:
Fixes #402

You can now chat with other users privately in the room group chat. We call this _whispering_, thanks to _James-Baloyi_ 

Previously, users were mentioned by putting an `@` before the username and selecting the user from the list. 

To whisper, put `#@` before selecting the user from the user list. This will only send the message to usernames that were mentioned using a `#@` before the username.

A working demo:

https://user-images.githubusercontent.com/38838675/109415336-d5fe2980-79d9-11eb-81b8-78a490a48e66.mp4

### DIFF
--- a/kofta/src/vscode-webview/modules/room-chat/RoomChatList.tsx
+++ b/kofta/src/vscode-webview/modules/room-chat/RoomChatList.tsx
@@ -38,86 +38,101 @@ export const RoomChatList: React.FC<ChatListProps> = ({}) => {
       ) : null}
       <div className={`pb-6`} />
       {messages.map((m) => (
-        <div className="flex items-center" key={m.id}>
+        <div className="flex flex-col" key={m.id}>
+          {/* Whisper label */}
+          {m.isWhisper ? (
+            <p className="mb-0 text-xs text-gray-400 px-2 bg-simple-gray-3a w-16 rounded-t mt-1 text-center">
+              Whisper
+            </p>
+          ) : null}
+
           <div
-            className={`py-1 block break-words max-w-full items-start flex-1`}
-            key={m.id}
+            className={`flex items-center px-1 ${
+              m.isWhisper
+                ? "bg-simple-gray-3a  py-1 rounded-b-lg rounded-tr-lg mb-1"
+                : ""
+            }`}
           >
-            <span className={`pr-2`}>
-              <Avatar size={20} src={m.avatarUrl} className="inline" />
-            </span>
-
-            <button
-              onClick={() => {
-                setProfileId(m.userId);
-                setMessagetobedeleted(
-                  (me?.id === m.userId ||
-                    iAmCreator ||
-                    (iAmMod && room?.creatorId !== m.userId)) &&
-                    !m.deleted
-                    ? m
-                    : null
-                );
-              }}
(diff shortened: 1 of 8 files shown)


kofta/src/vscode-webview/utils/createChatMessage.ts
```diff
@@ -35,5 +46,25 @@ export const createChatMessage = (message: string, mentions: BaseUser[]) => {
...
+  console.log({
+    tokens,
+    whisperedTo: roomUsers
+      .filter((u) =>
+        whisperedToUsernames
+          .map((u) => u?.toLowerCase())
+          .includes(u.username?.toLowerCase())
+      )
+      .map((u) => u.id),
+  });
```
benawad: could you remove this console.log

kofta/src/vscode-webview/modules/room-chat/RoomChatList.tsx
```diff
@@ -38,86 +38,101 @@ export const RoomChatList: React.FC<ChatListProps> = ({}) => {
...
-            </button>
+              <span className={`pr-2`}>
+                <Avatar size={20} src={m.avatarUrl} className="inline" />
+              </span>
 
-            <span className={`mr-1`}>: </span>
+              <button
+                onClick={() => {
+                  setProfileId(m.userId);
+                  setMessagetobedeleted(
```
benawad: camel case?

nadirabbas (commented):
is the Kousa side all okay?

kousa/lib/gen/room_chat.ex
```diff
@@ -59,13 +59,14 @@ defmodule Kousa.Gen.RoomChat do
...
-  def handle_cast({:new_msg, user_id, msg}, %State{} = state) do
+  def handle_cast({:new_msg, user_id, msg, whispered_to}, %State{} = state) do
     last_timestamp = Map.get(state.last_message_map, user_id)
     {_, seconds, _} = :os.timestamp()
 
     if not Map.has_key?(state.ban_map, user_id) and
-         (is_nil(last_timestamp) or seconds - last_timestamp > 0) do
-      ws_fan(state.users, :chat, %{
+      (is_nil(last_timestamp) or seconds - last_timestamp > 0) do
+      users = if whispered_to != [], do: Enum.filter(state.users, fn uid -> Enum.member?([user_id] ++ whispered_to, uid) end), else: state.users
```
benawad: better way to add element to array: `[user_id | whispered_to]`

benawad: also do it before you iterate

nadirabbas: whispered_to is an array itself. two arrays can be merged by pip character? or is it just like pushing an element to an array?

nadirabbas: Oh Okay got it @benawad

PR #484, opened 2021-02-28
feat: chat timestamps

### LINKED ISSUE #186: Add timestamps to chat messages
It'll be helpful to have timestamps for chat messages

nadirabbas:
Fixes #186

Gets UTC from server-side and converts to local.

image


kofta/package.json
```diff
@@ -11,6 +11,7 @@
     "@types/jest": "^26.0.15",
     "@types/node": "^12.0.0",
     "@types/react-dom": "^16.9.8",
+    "dateformat": "^4.5.1",
```
benawad: can you use date-fns so we use the same date lib for everything

nadirabbas: Okay sure.

kofta/src/vscode-webview/utils/format.ts
```diff
@@ -0,0 +1,7 @@
+var { format: dfsFormat } = require("date-fns");
```
benawad: use import syntax

benawad:
conflicts

nadirabbas:
oh they weren't there a minute ago lol

PR #532, opened 2021-03-02
Increased Websocket Timeout

ermalsh:
From 4 seconds to 15 seconds.

### DIFF
--- a/kofta/src/createWebsocket.ts
+++ b/kofta/src/createWebsocket.ts
@@ -29,9 +29,39 @@ export const createWebSocket = () => {
 
   useSocketStatus.getState().setStatus("connecting");
 
-  ws = new ReconnectingWebSocket(apiBaseUrl.replace("http", "ws") + "/socket");
-  const { accessToken, refreshToken } = useTokenStore.getState();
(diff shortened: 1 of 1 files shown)


benawad:
this timeout is fine, we need to increase the one for ReconnectingWebsocket

ermalsh:
> this timeout is fine, we need to increase the one for ReconnectingWebsocket

Gotcha, any preference to the increase amount? Default is 4000ms.

benawad:
do 15 seconds

kofta/src/createWebsocket.ts
```diff
@@ -29,9 +29,39 @@ export const createWebSocket = () => {
...
+    maxRetries: Infinity,
+    maxEnqueuedMessages: Infinity,
+    startClosed: false,
+    debug: false,
+  */
+  const wsOpts = {
+    connectionTimeout: 15000 //15 seconds
+  };
 
+  ws = new ReconnectingWebSocket(apiBaseUrl.replace("http", "ws") + "/socket", [], wsOpts);
```
benawad, line 62: just pass the option directly in and remove the docs

PR #541, opened 2021-03-02
📓 Added HUN translation

KristofKekesi:


ermalsh:
@KristofKekesi Looks good! Thank you for doing this. You may remove the _TODO comment if you'd like as well in this translation file.

PR #542, opened 2021-03-02
chore: added nvmrc on kofta

joshrezende:
help to maintain everyone with the same node version if using NVM.


benawad:
use 14 and this shouldn't have a yarn.lock

PR #547, opened 2021-03-03
feat: room description (pinned messages)

### LINKED ISSUE #192: Pinned messages for chat
have the ability for mods to pin certain messages, so that they will be stored. Any new members that join the room will be able to view the pinned chat messages. I don't know how viable this would be since you are likely not storing any chat information permanently as of now, but just a suggestion. It could be used to pin important links or a message describing what the room is talking about.

### LINKED ISSUE #518: Add room description (basically pinned message)

nadirabbas:
fixes #518 and #192 

working demo:

https://user-images.githubusercontent.com/38838675/109880664-09221080-7c99-11eb-9d17-bebf378ee537.mp4


kousa/lib/business-logic/room_logic.ex
```diff
@@ -180,20 +180,26 @@ defmodule Kousa.BL.Room do
...
-  end
-
-  def rename_room(user_id, new_name) do
-    with {:ok, room_id} <- Data.User.tuple_get_current_room_id(user_id),
-         {1, _} <- Data.Room.update_name(user_id, new_name) do
-      nil
-      RegUtils.lookup_and_cast(Gen.RoomSession, room_id, {:new_room_name, new_name})
+  def edit_room(user_id, new_name, new_description, is_private) do
+
+    cond do
```
ityonemo: these should be changeset validations

nadirabbas:
@ityonemo Thanks for the feedback. Could you kindly check if it's ok now?

kofta/src/vscode-webview/Routes.tsx
```diff
@@ -393,6 +402,11 @@ export const Routes: React.FC<RoutesProps> = () => {
 				setPublicRooms(() => ({ publicRooms, nextCursor }));
 			},
 			join_room_done: (d) => {
+				// Auto open chat to show description and if mobile
+				if (shouldBeSidebar) {
```
benawad: you'll need to make this a ref

kofta/src/vscode-webview/Routes.tsx
```diff
@@ -393,6 +402,11 @@ export const Routes: React.FC<RoutesProps> = () => {
...
+					useRoomChatStore.getState().toggleOpen();
```
benawad: someone could have chat from previous room already open

kofta/src/vscode-webview/modules/room-chat/RoomDescription.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useState } from "react";
+import { useCurrentRoomStore } from "../../../webrtc/stores/useCurrentRoomStore";
+import { truncate } from "../../utils/truncate";
+
+interface RoomDescriptionProps {}
+
+export const RoomDescription: React.FC<RoomDescriptionProps> = () => {
+	const { currentRoom } = useCurrentRoomStore();
+	const [expanded, setExpanded] = useState(false);
+	const MAX_COLLAPSED_CHARACTERS = 100;
```
benawad: this is a minor thing, but you can move this outside the render

kousa/lib/beef/room.ex
```diff
@@ -25,6 +27,7 @@ defmodule Beef.Room do
   @primary_key {:id, :binary_id, []}
   schema "rooms" do
     field(:name, :string)
+    field(:description, :string)
```
benawad: change this to `field(:description, :string, default: "")`

kousa/lib/beef/room.ex
```diff
@@ -42,9 +45,22 @@ defmodule Beef.Room do
...
+
+  @spec edit_changeset(
+          {map, map} | %{:__struct__ => atom | %{__changeset__: map}, optional(atom) => any},
+          :invalid | %{optional(:__struct__) => none, optional(atom | binary) => any}
+        ) :: Ecto.Changeset.t()
+  def edit_changeset(room, attrs) do
+    room
+    |> cast(attrs, [:id, :name, :isPrivate, :description])
+    |> validate_required([:name])
+    |> validate_length(:name, min: 2, max: 255)
```
benawad: new max is `60`

kofta/src/vscode-webview/components/CreateRoomModal.tsx
```diff
@@ -11,99 +11,125 @@ import { InputField } from "./form-fields/InputField";
...
-                <option value="private" className={`bg-simple-gray-3c`}>
-                  private
-                </option>
-              </select>
-            </div>
+					return errors;
+				}}
+				onSubmit={async ({ name, privacy, description }) => {
+					if (edit) {
+						wsend({
```
benawad: switch this to use `wsFetch`, see what the modal on `staging` looks like for example

kousa/lib/socket_handler.ex
```diff
@@ -361,8 +362,8 @@ defmodule Kousa.SocketHandler do
     {:ok, state}
   end
 
-  def handler("edit_room_name", %{"name" => name}, state) do
```
benawad, line 366: because `DogeHouse` runs in "prod", you can't rename/delete any handlers (because old clients will still call these). Add a comment about deprecating them then add new handlers

kousa/lib/utils/errors.ex
```diff
@@ -1,8 +1,9 @@
 defmodule Kousa.Errors do
-  def changeset_to_first_err_message(%{errors: [{_, {message, values}}]}) do
-    Enum.reduce(values, message, fn {k, v}, acc ->
+  def changeset_to_first_err_message(%{errors: [{field, {message, values}}]}) do
```
benawad: create a separate function for return field name + error, sometimes I just want error

kousa/priv/repo/migrations/20210303204515_room_description.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Beef.Repo.Migrations.RoomDescription do
+  use Ecto.Migration
+
+  def change do
+    alter table(:rooms) do
+      add :description, :text, default: ""
```
benawad: also set `null: false`

kofta/src/vscode-webview/components/CreateRoomModal.tsx
```diff
@@ -11,99 +11,125 @@ import { InputField } from "./form-fields/InputField";
...
+				description?: string;
+			}>
+				initialValues={{
+					name: currentName || "",
+					description: currentDescription || "",
+					privacy: isPrivate ? "private" : "public",
+				}}
+				validateOnChange={false}
+				validateOnBlur={false}
+				validate={({ name }) => {
```
benawad: do same validation you do on the backend on the frontend

benawad (changes requested):

kofta/src/vscode-webview/Routes.tsx
nadirabbas: yeah, I tried that but seemed like once they leave a room, it is closed.

kousa/lib/socket_handler.ex
nadirabbas: so a comment would be sufficient? Won't the old clients still fail?

nadirabbas:
@benawad I can no longer find room_data.ex was it removed?

nadirabbas:
@benawad getting `RoomBlock.get_blocked_users` is undefined. I think there were huge refactoring changes and now I am lost . lol

benawad:
`room_data` is now `rooms`

`get_blocked_users` should still exist in same location

nadirabbas:
image

@benawad  this happens when i open the settings modal in room. i have the latest staging changes. can you check if this also happens on your side?

benawad:
should be `RoomBlock` not `RoomsBlock`

nadirabbas:
@benawad that means this error is also in staging

nadirabbas:
image
image

@benawad this is happening in all cases, whether its `RoomBlock`, `RoomBlocks`. I cannot even found the `def` of `get_blocked_users` in `beef/room_block`. The def exists in the `room_block_data` which I think is deleted from `staging`

Also tried `alias Beef.RoomBlock`

nadirabbas:
May be I should copy `get_blocked_users` from `room_block_data` to `room_block`

ityonemo:
I would search for those function defs. The "data layer" modules have been moved to Beef.TablePlural module, which is more idiomatic elixir.

nadirabbas:
@ityonemo I couldn't find the `get_blocked_users` function in `beef/room_block` which is being accessed in some other location and throwing an undefined error. Let me know if you find it.

ityonemo:
I think I messed up the refactor =( I'll add to this pr and restore the funciton.

nadirabbas:
@ityonemo I already restored it and it works, I'll push it on this or. put it in beef/room_blocks

ityonemo:
so roomblocks and userblocks are the only things I haven't managed to get to the refactor yet, so they should still go in through the old data-layer modules.

ityonemo:
yeah as long as it works it's fine.  I'm going to rebase to existing staging before dropping the last one in.

benawad:
ok it's fixed in staging now @nadirabbas

nadirabbas:
@benawad Ready for review.

nadirabbas:
@benawad not sure what is happening, i have resolved conflicts like 5 times 😂😂😂 Every 5 minutes I have more conflicts

kofta/src/app/Routes.tsx
```diff
@@ -393,6 +404,11 @@ export const Routes: React.FC<RoutesProps> = () => {
 				setPublicRooms(() => ({ publicRooms, nextCursor }));
 			},
 			join_room_done: (d) => {
+				// Auto open chat to show description and if mobile
+				if (shouldBeSidebarRef) {
```
benawad: needs `.current`

kofta/src/app/components/CreateRoomModal.tsx
```diff
@@ -25,31 +33,36 @@ export const CreateRoomModal: React.FC<CreateRoomModalProps> = ({
...
 
-					if (name.length < 2) {
+					if (name.length < 2 || name.length > 60) {
 						return {
-							name: "min length 2",
+							name: "must be between 2 to 60 characters long",
+						};
+					} else if (description.length > 500) {
+						return {
+							description: "min length 500",
```
benawad: max, not min

kousa/lib/business-logic/room_logic.ex
```diff
@@ -189,10 +189,12 @@ defmodule Kousa.BL.Room do
     })
   end
 
+  # @deprecated
   def rename_room(_user_id, new_name) when byte_size(new_name) > 255 do
     {:error, "name needs to be less than 255 characters"}
   end
 
+  # @deprecated
   def rename_room(user_id, new_name) do
```
benawad: handlers are a public interface so they need to be deprecated but all other elixir functions can be safely edited/deleted

benawad (changes requested):

nadirabbas:
Thanks for taking the time to actually fix my mistakes @benawad learning a lot.

benawad:
for sure, I appreciate the help!

PR #551, opened 2021-03-04
fix: show user modal if no longer in the room

nadirabbas:
Clicking on a user in chat doesn't open their profile if they have left the room.
This PR adds the ability to make a request to API to get user profile ONLY IF they have left the room.


nadirabbas:
@benawad we have this: `Cannot find module: '@material-ui/core/styles'. Make sure this package is installed.` idk why

kofta/src/vscode-webview/types.ts
```diff
@@ -1,58 +1,58 @@
...
+	muteMap: Record<string, boolean>;
+	activeSpeakerMap: Record<string, boolean>;
+	autoSpeaker: boolean;
 };
 
 export type WsParam = {
-  op: string;
-  d: any;
+	op: string;
+	d: any;
```
benawad: keep og spacing

kofta/src/vscode-webview/atoms.ts
```diff
@@ -3,88 +3,101 @@ import { useCurrentRoomStore } from "../webrtc/stores/useCurrentRoomStore";
 import { Room, BaseUser, UserWithFollowInfo } from "./types";
 
 const createSetter = <T>(a: WritableAtom<T, any>) =>
-  atom(null, (get, set, fn: (x: T) => T) => {
-    set(a, typeof fn === "function" ? fn(get(a)) : fn);
-  });
+	atom(null, (get, set, fn: (x: T) => T) => {
```
benawad: use og spacing

(is prettier hook doing this?)

kofta/src/vscode-webview/modules/room-chat/ProfileModalFetcher.tsx
```diff
@@ -1,53 +1,68 @@
...
-  );
+	const profileFromRoom: RoomUser | undefined = room?.users.find((x) =>
+		[x.id, x.username].includes(userId)
+	);
+
+	const profile = profileFromRoom || userProfile;
+
+	useEffect(() => {
+		profileFromRoom ||
+			wsend({ op: "get_user_profile", d: { userId, userIdType } });
```
benawad: this is a clunky pattern, I've switched to using `react-query` to fetch stuff see `PublicRoomsQuery` as an example  https://github.com/benawad/dogehouse/blob/staging/kofta/src/vscode-webview/pages/Home.tsx#L42

(you'll also use an f_handler for this on the backend)

kofta/src/vscode-webview/atoms.ts
nadirabbas: @benawad its actually the pre-hook commit. you have the run `npm install` in the root to enable it for yourself too,

kofta/src/vscode-webview/modules/room-chat/ProfileModalFetcher.tsx
nadirabbas: Even if I use `react-query`, I can still broadcast events to other people right?

kofta/src/vscode-webview/atoms.ts
```diff
@@ -3,88 +3,101 @@ import { useCurrentRoomStore } from "../webrtc/stores/useCurrentRoomStore";
...
+	}
+
+	const isCreator = me.id === room.creatorId;
+
+	return {
+		isCreator,
+		isMod,
+		isSpeaker,
+		canSpeak: isCreator || isSpeaker,
+	};
```
nadirabbas: hook changed spacing on this file @benawad

nadirabbas:
@benawad Now it's using `wsFetch`. btw all these spacing-related changes are due to the pre-commit hook. I just installed it.

kofta/src/vscode-webview/atoms.ts
benawad: shouldn't the hook only format files you touch? Also can we configure it to use 2 spaces

nadirabbas: @benawad it does it project wide

kofta/src/vscode-webview/modules/room-chat/ProfileModalFetcher.tsx
benawad: yeah it just replaces wsend

kofta/src/vscode-webview/atoms.ts
benawad: we should move husky to `kofta`, the point of the hook is to style other contributors commits that don't know the format yet, but I doubt many of them will `npm i` at the root

PR #564, opened 2021-03-05
remove [emote soon] text

xkral-tr:


nadirabbas:
why are you removing it?

xkral-tr:
> why are you removing it?

because they added it or not

PR #566, opened 2021-03-05
PT-PT translation + fix typos

miguelcarreiro:
This PR includes:
- European portuguese (pt-pt) translation;
- Fix typos in translation keys.

@benawad In the language selector, I suggest that Brazilian Portuguese be represented as `pt-br` (can change that if needed), since there are now two portuguese dialects (`pt-br` + `pt-pt`).


benawad:
yeah if you could change that, that would be great

miguelcarreiro:
@benawad Done

Issue #579, opened 2021-03-06
Clicking on "ask to speak" on frontend issues function RoomsPermission.ask_to_speak/2 is undefined (module RoomsPermission is not available)


Choongkyu:
**Describe the bug**

**To Reproduce**
1. Create room
2. In another browser, either different type, or the same type in incognito mode to avoid cookie, login as another user and join room created in step 1
3. Click on megaphone icon to ask to speak
4. See error `function RoomsPermission.ask_to_speak/2 is undefined (module RoomsPermission is not available)`

**Expected behavior**
No error

**What device are you on?**
macos big sur

This is due to bug introduced in https://github.com/benawad/dogehouse/pull/543

shmuelhizmi:
+1

PR #581, opened 2021-03-06
Fix bug #579

Choongkyu:
Correct ask_to_speak bug due to calling undefined RoomsPermission module to using proper module


benawad:
mb, didn't see this until I just pushed

Issue #578, opened 2021-03-06
Record sessions


Mudrank:
Since we discussed this in the weekly contributor meeting I thought This should have a feature request

shmuelhizmi:
if #585 will pass support for audio recording will be added

benawad:
0. I think we want to do the recording server side, MediaRecorder doesn't support iOS Safari https://caniuse.com/mediarecorder (this also needs to be thought out of how it works with deploys so a recording is not interrupted)
1. Only the creator of the room can initiate a recording
2. All current speakers need to confirm they are ok with it being recorded
3. It should be obvious to everyone in the room its being recorded and when someone joins to speak they should be prompted to confirm they are ok with this
4. The recording will be saved to Google Cloud probably
5. The recording will show up on the profile of everyone who spoke and the creator of the room can set the privacy of it to public/private (if set to private can the other speakers still see it??)

JammSpread:
What if it can be configured that speakers who don't decide to be recorded are not recorded?

shmuelhizmi:
I think taking in consideration doge-house user demographic, user will have no problem just using their PC to record the session and so being able to record the session directly from the app is just a matter of convenience, therefore preventing users from recording other users without thier consent isn't a matter of security but a matter responsibility (you don't want users to use your app feature to record users other users without indication that they are being recorded), taking that in consideration I want to propose a simpler solution that is much easier to implement, require less overhead from the users but is also just as responsible.

The solution is just showing a red dot below the user that is recording, (and maybe also send a message is the chat that the user has start recording) this will late all user understand that they are being recorded, and let moderators who think that the user should not be allowed to record the session kick him out.


As for cloud recording even though this can have lots of cool benefits, this will give us lots of troubles, cloud recording is complicated and that is way companies like zoom defaults to local recording ,zoom cloud recording require up to 72 hours of processing and is only available for paid users, this will not be the case with as since we only work with audio but still cloud recording is much more resource demanding compering to a regular media-server.

benawad:
Cloud recording is more complex, but needed for this project. One of the goals of the project is to be a platform podcasters can host their podcasts on.

ubrg:
@benawad 

Would you like to do a session record of the audio-only, or something similar to what Twitch does with its replay feature (video + chat + maybe other stuff... are replayed as if it was the actual session)?

[...]

PR #585, opened 2021-03-06
support for recording audio

shmuelhizmi:
support for recording room audio


shmuelhizmi:
closing due to a bug (you are not self-recording)

shmuelhizmi:
I'm also having troubles since I merge staging to my branch due to #579 
@benawad

shmuelhizmi:
I fixed the bug, a few notes:
1 - currently there is no icon for the recording button.
2 - users volume is not applied to the recording, I can fix that by setting the volume of the users inside the track and not in the audio element like it is now, I just wasn't sure if there is a specific reason we set the volume in the element and not the track.
3 - currently I just make the browser download the audio file the second the recording is done, do we prefer to ask the user first?
4 - do we want the recording file name to say something specific? maybe `[the_room_name].mp3` ?

kofta/src/app/components/BottomVoiceControl.tsx
```diff
@@ -177,6 +179,34 @@ export const BottomVoiceControl: React.FC<BottomVoiceControlProps> = ({
...
+
+		if (recorder.isRecording) {
+      buttons.push(
+        <button
+          className={buttonStyle}
+          key="stop-recording"
+          onClick={() => {
+            recorder.toggle();
+          }}
+          title="Stop Recording"
```
gitaumoses4: The title could be translated as well.

kofta/src/app/components/BottomVoiceControl.tsx
```diff
@@ -177,6 +179,34 @@ export const BottomVoiceControl: React.FC<BottomVoiceControlProps> = ({
...
+      );
+    } else {
+      buttons.push(
+        <button
+          className={buttonStyle}
+          key="start-recording"
+          onClick={() => {
+            recorder.toggle();
+          }}
+          title="Start Recording"
```
gitaumoses4: Same here.

benawad:
see  #578 for notes, this may be hard to do atm

cdbattags:
I think we can start [here](https://github.com/ethand91/mediasoup3-record-demo) as an example?

https://github.com/ethand91/mediasoup3-record-demo

Issue #553, opened 2021-03-05
Adjust positioning of login buttons and footer links


andrewroblesdev:
When viewed from a desktop, login buttons and footer links should be the same width of the text above it in order to be more visually appealing.

andrewroblesdev:
Quick mockup to illustrate what I mean:
![Screenshot 2021-03-05 083444](https://user-images.githubusercontent.com/70864612/110144913-bd04c800-7d8d-11eb-9b33-e530e522b068.png)

andrewrobles:
Not sure if there issues need to be approved before anyone starts working on them. But I would like to work on this issue

andrewroblesdev:
Not sure if issues need to be approved before anyone starts working on them. But I would like to work on this issue

benawad:
sounds good, go for it!

PR #586, opened 2021-03-06
Issue #553 - Reposition login buttons and links on desktop mode

andrewroblesdev:
The login buttons were made wider and the links were spaced out to match the alignment of the rest of the content on the login page. See attached screenshot to result.
Screen Shot 2021-03-06 at 5 07 10 AM

### DIFF
--- a/kofta/src/app/components/Footer.tsx
+++ b/kofta/src/app/components/Footer.tsx
@@ -11,7 +11,7 @@ export const Footer: React.FC<FooterProps> = ({ isLogin }) => {
 	const { t } = useTypeSafeTranslation();
 
 	return (
-		<div className={`justify-around flex text-center`}>
+		<div className={`justify-between flex text-center`}>
 			{isLogin ? (
 				<RegularAnchor href="https://www.youtube.com/watch?v=hy-EhJ_tTQo">
 					{t("footer.link_1")}


kofta/src/app/components/Button.tsx
```diff
@@ -87,7 +87,7 @@ export const Button: React.FC<
             variant === "small"
               ? "py-1 px-2 w-max"
               : variant === "slim"
-              ? "max-w-md ml-auto mr-auto py-2.5 px-12"
+              ? "max-w-2xl ml-auto mr-auto py-2.5 px-12"
```
benawad: I like that the buttons are not full width on desktop atm

kofta/src/app/components/Footer.tsx
```diff
@@ -11,7 +11,7 @@ export const Footer: React.FC<FooterProps> = ({ isLogin }) => {
 	const { t } = useTypeSafeTranslation();
 
 	return (
-		<div className={`justify-around flex text-center`}>
+		<div className={`justify-between flex text-center`}>
```
benawad, line 14: will this smash the footer against the side on mobile?

kofta/src/app/components/Button.tsx
andrewroblesdev: Sounds good, I will remove this change from the pull request

kofta/src/app/components/Footer.tsx
andrewroblesdev: No I don't think so. Looks almost identical to how it was before on mobile: below is the resulting screenshot. If you don't like it, I can revert this too no problem.
Screen Shot 2021-03-06 at 5 54 02 AM

andrewroblesdev:
Update: the width of the buttons has been reverted. Attached is the screenshot result of the most recent state of the pull request:
Screen Shot 2021-03-06 at 6 02 07 AM

PR #594, opened 2021-03-06
Added Additional EN Translation for Modules

ermalsh:
- ScheduledRooms Module Translations
- RoomChat Module Translations

Will need to add Strings for these new additional translations to the other translation JSON files. Will fall back to EN if translation string is not found for selected language.


benawad:
If you add more keys to `en`, next time run `npm run i18` after doing that and only use `useTypeSafeTranslation` hook

PR #595, opened 2021-03-06
multiple bug fixes

### LINKED ISSUE #593: can't edit room description
at the moment you cannot edit the room description once you have created it. I had added support for it but I guess it got lost in all the conflicts resolving.  will repush it

image

nadirabbas:
Fixes #593 

- set new unread messages to false on window focus
- make room description editable after room creation
- retain room description & bio indentation


kofta/src/app/modules/room-chat/RoomChatList.tsx
```diff
@@ -33,6 +33,7 @@ export const RoomChatList: React.FC<ChatListProps> = ({}) => {
 	// Only scroll into view if not manually scrolled to top
 	useEffect(() => {
 		isRoomChatScrolledToTop || bottomRef.current?.scrollIntoView();
+		window.addEventListener("focus", () => setNewUnreadMessages(false));
```
benawad: you need to put this in a different useEffect or you will create hundreds of listeners, you should also stop listening on unmount

benawad:
wait, what is the point of `set new unread messages to false on window focus`

nadirabbas:
@benawad  so if you have `chat` opened, but you are on a different tab. Somebody mentions you, there is a red dot visible on the chat square on the `BottomVoiceControl`. Even if you switch to the Doghouse window, that red dot won't go away unless you close the chat and reopen it. 

This PR hides the red dot when `window` is focused.

benawad:
shouldn't you make the red dot only appear if chat is closed instead?

nadirabbas:
@benawad  yeah, that's actually correct, will fix it.

PR #596, opened 2021-03-06
feat: notifications

nadirabbas:
- [x] Add notifications table & schemas
- [x] Notify users on notifications
- [x] Add follow notification 
- [x] Add notification toast
- [x] Set notification `is_read` to `true` when read
- [ ] Add notifications page
- [ ] Add indication for unread notifications


nadirabbas:
@benawad can't use `useTypeSafeTranslation` with dynamic strings.

kousa/lib/beef/notifications.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Beef.Notifications do
+  alias Beef.Schemas.Notification
+
+  def insert(data) do
+    %Notification{}
+    |> Notification.insert_changeset(data)
+    |> Beef.Repo.insert(on_conflict: :nothing, returning: true)
```
nadirabbas, line 9: @benawad  could you help me when u get the time? How can I load assoc directly when a record is inserted? searched a lot but couldn't figure it out :(

kousa/priv/repo/migrations/20210305231415_notifications_table.exs
```diff
@@ -0,0 +1,14 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:notifications, primary_key: false) do
+      add :id, :uuid, primary_key: true, default: fragment("uuid_generate_v4()")
+      add :type, :string, null: false
+      add :is_read, :boolean, default: false
+      add :user_id, references(:users, on_delete: :delete_all, type: :uuid), null: false, primary_key: true
+      add :notifier_id, references(:users, on_delete: :delete_all, type: :uuid), null: true, primary_key: true
+      timestamps()
```
benawad, line 11: we switched to `:utc_datetime_usec` do:

```
      add :inserted_at, :utc_datetime_usec, null: false, default: fragment("now()")
      add :updated_at, :utc_datetime_usec, null: false, default: fragment("now()")
```
instead of `timestamps()`

kousa/lib/beef/schemas/notification.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  @primary_key {:id, :binary_id, []}
+  schema "notifications" do
+    field(:type, :string)
+    field(:is_read, :boolean)
+
+    belongs_to(:user, User, foreign_key: :user_id, type: :binary_id)
+    belongs_to(:notifier, User, foreign_key: :notifier_id, type: :binary_id)
+
+    timestamps()
```
benawad, line 26: look at `ScheduledRoom` schema for how to do inserted_at instead of using `timestamps()`

kousa/lib/beef/notifications.ex
benawad: just fetch the user by id in the business logic fn

benawad:
you can add a new circle button on the home page (like scheduled rooms) to go to notifications page

benawad:
we probably also want to show the # of unread notifications, but that's going to be a little tricky

nadirabbas:
image

nadirabbas:
@benawad 

https://user-images.githubusercontent.com/38838675/110243527-9658b400-7f7c-11eb-99d9-82516ccacf04.mp4

nadirabbas:
@benawad should I only show unread notifications in the notifications page? that way we can delete read notifications.

benawad:
no don't worry about that, show all of them (paginated of course), we can always add a cron job later to delete really old notifs.

add a follow back button (only if you don't already follow them)

PR #608, opened 2021-03-07
Enhancement/#600 display according to users intl selection

### LINKED ISSUE #600: Kofta: Display dates according to the users 'cultural convention'
Reading American dates is very confusing, they should be displayed according to the language the website is beeing accessed in.
grafik
Additionally automatically switching between 12H and 24H format would be helpful.

ermalsh:
Closes #600 
Added Intl date formatter for dates
- Changes date format based on language selection
- Added "commons.formattedIntlDate" as TranslationKey
- npm run i18

Use:
`t("common.formattedIntlDate", "dateString")`


benawad:
It looks like it's showing seconds
image

I only want hr/min

Issue #617, opened 2021-03-07
Kousa and Shawarma can only connect to RabbitMQ instance if they are running after RabbitMQ is ready and listening


Choongkyu:
**Describe the bug**
Kousa and Shawarma, if running before RabbitMQ instance or anytime before the instance is ready, cannot connect to RabbitMQ.  They have no retry mechanism or any other opportunity to connect to RabbitMQ instance.

**To Reproduce**
1. Run Kousa or Shawarma
2. Run RabbitMQ server
3. Both Kousa and Shawarma will issue an error but remain running without being able to connect to RabbitMQ

**Expected behavior**
Kousa and Shawarma should both be trying to reconnect to RabbitMQ indefinitely until they're successful.  This becomes especially salient when deploying with any sort of orchestration like Kubernetes or Docker Compose/Swarm, where there is no guarantee of running order and either Kousa or Shawarma.

**What device are you on?**
macos big sur

PR #628, opened 2021-03-07
Address issue #617 [shawarma, kousa] retry on rabbit connect failure

Choongkyu:


shawarma/src/index.ts
```diff
@@ -14,366 +14,379 @@ const log = debugModule("shawarma:index");
...
-    router: Router;
-  }[];
-  try {
-    workers = await startMediasoup();
-  } catch (err) {
-    console.log(err);
-    throw err;
-  }
-  let workerIdx = 0;
+	if (process.env.SENTRY_DNS) {
```
benawad: match OG spacing, tab width 2 spaces, should be in prettier in root

kousa/lib/gen/voice_rabbit.ex
```diff
@@ -25,19 +25,26 @@ defmodule Kousa.Gen.VoiceRabbit do
   @send_queue "shawarma_queue"
   @receive_exchange "kousa_exchange"
   @receive_queue "kousa_queue"
+  @retry_interval 15_000
```
benawad: I want the interval to be 5 seconds

kousa/lib/gen/voice_rabbit.ex
```diff
@@ -25,19 +25,26 @@ defmodule Kousa.Gen.VoiceRabbit do
...
+        queue_to_consume = @receive_queue <> opts.id
+        IO.puts("queue_to_consume: " <> queue_to_consume)
+        # Register the GenServer process as a consumer
+        {:ok, _consumer_tag} = Basic.consume(chan, queue_to_consume, nil, no_ack: true)
+        {:ok, %State{chan: chan, id: opts.id}}
+      {:error, _} ->
+        IO.puts("[Rabbit] error on connecting to server: #{host}")
+        :timer.sleep(@retry_interval)
+        init(opts)
+    end
```
benawad: does this only work on initial startup? What about when the connection gets dropped after init?

Choongkyu: Yes. I should have mentioned that this is only for initial startup to address an issue I was having with syncing containers. I can extend this to address dropped connections as well.

kousa/lib/gen/voice_rabbit.ex
Choongkyu: thanks! I'll change this.

shawarma/src/index.ts
Choongkyu: Thanks for catching this subtle but annoying change! I'm actually a bit puzzled by this since I'm seeing a tab width of 2 spaces on vs code and prettier is on and my settings seem to check out. I'll dig more into this.

Choongkyu: I just found this https://github.com/github/VisualStudio/issues/1524 , which seems to explain the tab width discrepancy between vs code and `git diff <hash>`.  Adding `useTabs: false` to the prettier config fixes this for me. Would it be OK to add that property? Incidentally, this aligns with my preference for spaces over tabs, which might explain why I've never come across this before. That still doesn't explain why no other contributor on this repo has seen this issue, though. 🤷🏻

kousa/lib/gen/voice_rabbit.ex
Choongkyu: I'm not familiar with Elixir well enough to work with the supervisor pattern, but I believe that's what would need to be used to restart the RabbitMQ-related processes.  Prior to the changes in this PR, the application completely crashes if no RabbitMQ connection is established.  After this PR, the application completely crashes if no RabbitMQ connection established except for during initial startup.  Can the reconnect logic be added in another PR for the time being? I did add retry logic to Shawarma though that handles disconnects even after init.

Choongkyu: I was able to update the logic to enable reconnecting with amqp after all.

benawad:
why is it showing so many people's commits?

Choongkyu:
> why is it showing so many people's commits?

good question. I'll close this pull request and start a fresh one. I had rebased and after doing so, somehow my local copy fell behind the remote one, which doesn't make sense to me.

PR #638, opened 2021-03-07
feat(locales): improve locale support

Johannes-Andersen:


Johannes-Andersen:
Humm.. seems like formatting is off. Taking a look now

Edit: should be fixed now

PR #642, opened 2021-03-08
Added hebrew support

vi241:


ofsho:
Doesn't the pr autoclose when merged?

PR #647, opened 2021-03-08
Change All CRLF line endings to LF endings

Angshu31:
The ESLint was configures to only like LF endings but literally every file (even `.storybook/` files) had CRLF endings.

PR changes it to LF


vi241:
what is that?

PR #648, opened 2021-03-08
Hotkeys for toggle-mute, push-to-talk, and toggle-chat

andrewroblesdev:
Render keybind listeners for toggle-mute, push-to-talk, and toggle-chat to bottom voice control when user is in a room.

Default keybinds are:
- `Control+m` for toggle-mute
- `0` for push-to-talk
- `c` for toggle-chat

These keybinds can be changed in voice settings.


nadirabbas:
Should toggle-chat really be in voice settings? @benawad we should have a separate page only for keybinds, as we have for sound settings.

benawad:
we will probably rename this to keybind settings in new design

Issue #637, opened 2021-03-07
[New Design] Atomic Components Implementation v1


benawad:
<h1>Atomic Components</h1>

(do not add margin on any atomic components)

- [x] Button: https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A731
- [x] Tag: https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A680
- [x] IconHeading - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A681
- [x] ScheduledTime - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A686
- [x] UserAvatar - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A679
- [x] BoxedIcon - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A704
- [x] UserBadge - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A725
- [x] SearchBar - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=499%3A1619
- [x] SearchHistory - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=499%3A1697
- [x] SearchOverlay - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=499%3A1696
  - this is just the container
- [x] SettingsIcon - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=499%3A1773
- [x] BaseOverlay - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=499%3A1842
  - this will be used to make SettingsOverlay/NotificationsOverlay/Message

<h1>Compound Components</h1>

- [x] SearchResult - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=499%3A1747
  -  requires UserAvatar & ScheduledTime & SearchHistory ✅
- [x] MessageElement - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=499%3A2053
  - requires UserAvatar ✅
[...]

PR #649, opened 2021-03-08
Tag component for new-design

Jackbaude:
Apart of https://github.com/benawad/dogehouse/issues/637


Jackbaude:
Committed in the wrong dir

PR #660, opened 2021-03-08
ScheduledTime UI component

jee1mr:
Implement ScheduledTime component from the list here: https://github.com/benawad/dogehouse/issues/637

### DIFF
--- a/kibbeh/src/ui/ScheduledTime.tsx
+++ b/kibbeh/src/ui/ScheduledTime.tsx
@@ -1,0 +1,10 @@
+import React from "react";
+
+export const ScheduledTime: React.FC = ({ children }) => {
+  return (
+    <div className="text-primary-200 font-bold">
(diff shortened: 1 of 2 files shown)


kibbeh/src/ui/ScheduledTime.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+export const ScheduledTime: React.FC = ({ children }) => {
+  return (
+    <div className="text-primary-200 font-bold">
+      <div className="inline-block mr-2 w-2 h-2 rounded-full bg-primary-200"></div>
```
ajmnz: Background for the circle should be `bg-primary-300`

ajmnz (changes requested):

kibbeh/src/ui/ScheduledTime.tsx
jee1mr: ah right! my bad

ajmnz (approved):

PR #657, opened 2021-03-08
SearchHistory Component

mayconbenito:

PR #664, opened 2021-03-08
Add SearchBar component

kleo-xhindoli:
* Adds SearchBar component

~**Note:** The icon was actually 18x18 in the design but since there is no 18px spacing I had to use the closest value in the config which is 20px. Let me know if we need to add a 3.5 value to be 18px.~

**Note 2:** I'm using `React.ComponentPropsWithoutRef<'x'>` to extend base elements' props instead of `DetailedHTMLProps` as suggested [here](https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#wrappingmirroring). 
I suggest we use this going further as it's easier to write and understand what's happening.


kibbeh/src/ui/icons/SearchIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+interface SearchIconProps extends React.ComponentPropsWithoutRef<"svg"> {
+  className?: string;
+}
+
+export const SearchIcon: React.FC<SearchIconProps> = (props) => {
```
ajmnz: The svg has its own size already set. Try to use it like this https://github.com/benawad/dogehouse/blob/40f987073f21eb252d5c4258a919d486bf47712f/kibbeh/src/ui/SearchHistory.tsx#L3 so you don't run into sizing problems and also is consistent with the other search components.

ajmnz (changes requested):

kibbeh/tailwind.config.js
```diff
@@ -17,6 +17,17 @@ module.exports = {
     },
     colors: {
       button: "var(--color-button-text)",
+      gray: {
```
benawad: Use primary instead of gray

kibbeh/src/ui/icons/SearchIcon.tsx
kleo-xhindoli: I see new icons have been added. I replaced what I had previously with the existing `SmSolidSearch`. I had to change its stroke color so that we can override it from the parent component.

benawad (approved):

kibbeh/src/ui/SearchBar.tsx
```diff
@@ -0,0 +1,27 @@
...
+  inputClassName?: string;
+}
+
+export const SearchBar: React.FC<SearchBarProps> = ({
+  className,
+  inputClassName,
+  ...props
+}) => {
+  return (
+    <div
```
ajmnz, line 15: Make the component be a wrapper for the icon and the input, so we can ensure that the icon is `mr-4` apart from the input.

You can set the parent to `bg-gray-700 rounded-lg` (and remove those from the input), remove position absolute from the icon wrapper and set it to `mr-4` and remove `pl-7` from the input and set `bg-transparent`. Let me know if that makes sense, it has to match with https://github.com/benawad/dogehouse/pull/657

kibbeh/src/ui/SearchBar.tsx
```diff
@@ -0,0 +1,27 @@
...
+      className={`relative text-primary-300 focus-within:text-primary-100 w-full ${className}`}
```
ajmnz: `${className}` adds `undefined` to the class list. Might not be needed to add extra classes.

ajmnz (changes requested):

kibbeh/src/ui/SearchBar.tsx
kleo-xhindoli: I think passing through the className is useful when you want to add margin classes from the parent without having to add extra wrapper divs so I think we should leave it. I set the default values to be empty strings so that `undefined` doesn't get added to the classes list.

kibbeh/src/ui/SearchBar.tsx
kleo-xhindoli: Pushed some new changes. Can you check again?

kibbeh/src/ui/SearchBar.tsx
```diff
@@ -0,0 +1,27 @@
...
+}) => {
+  return (
+    <div
+      className={`w-full bg-primary-700 text-primary-300 focus-within:text-primary-100 rounded-lg ${className}`}
+    >
+      <div className="h-full mx-4 flex items-center pointer-events-none">
+        <SmSolidSearch />
+      </div>
+      <input
+        className={`w-full py-2 bg-transparent text-primary-100 placeholder-primary-300 focus:outline-none ${inputClassName}`}
```
ajmnz: Missed `pr-4` in the input

ajmnz (changes requested):

kibbeh/src/ui/SearchBar.tsx
kleo-xhindoli: ~I think it's not needed. There's no reason for the input not to expand to the full width.~

NVM it looks weird without it :P

ajmnz: It is needed so the text doesn't reach the end of the wrapper.

Screenshot 2021-03-08 at 23 57 43

vs

Screenshot 2021-03-08 at 23 58 17

ajmnz (approved):

PR #666, opened 2021-03-08
Boxed icon component

chancovsky:
# Boxed icon component

Figma reference: https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=448%3A704


kibbeh/src/ui/BoxedIcon.tsx
```diff
@@ -0,0 +1,18 @@
...
+import React, { HTMLAttributes, ReactNode } from "react";
+
+interface BoxedIconProps extends HTMLAttributes<HTMLElement> {
+  icon: ReactNode;
+}
+
+export const BoxedIcon: React.FC<BoxedIconProps> = ({ icon, ...props }) => {
+  return (
+    <div
+      className="bg-primary-700 justify-center items-center rounded"
```
ajmnz: Set the background to `gray-600` on hover and use `w-6 h-6` instead of inline styles https://github.com/benawad/dogehouse/blob/24d03500c4b4a84228cb897e9e30b8ac3304664d/kibbeh/tailwind.config.js#L42

Also `cursor-pointer`.

ajmnz (changes requested):

kibbeh/tailwind.config.js
```diff
@@ -27,6 +27,15 @@ module.exports = {
...
       },
+      gray: {
+        100: "var(--color-gray-100)",
+        200: "var(--color-gray-200)",
+        300: "var(--color-gray-300)",
+        600: "var(--color-gray-600)",
+        700: "var(--color-gray-700)",
+        800: "var(--color-gray-800)",
+        900: "var(--color-gray-900)",
+      },
```
benawad: use primary instead of gray

kibbeh/src/stories/BoxedIcon.story.tsx
```diff
@@ -0,0 +1,58 @@
...
+  title: "Boxed Icon",
+};
+
+const BoxedIconStory: Story = () => {
+  return (
+    <div className="flex flex-row">
+      <div className="m-1">
+        <BoxedIcon
+          icon={
+            <svg width="14" height="18" viewBox="0 0 14 18" fill="none">
```
benawad: Use the React components inside `src/icons` for icons

ajmnz:
Why did you change the icon sizes? That affected yours and other already finished components. Each icon is ready to be imported as is, with its width and height already set up.

chancovsky:
I changed the Icon sizes to normalize them against `font-size`.  This will allow for more consistent icon sizing across the application instead of having icons sized for specific use cases.  It also allows for more flexibility when using the icons and have a consistent way manage their size.

ajmnz:
> I changed the Icon sizes to normalize them against `font-size`. This will allow for more consistent icon sizing across the application instead of having icons sized for specific use cases. It also allows for more flexibility when using the icons and have a consistent way manage their size.

~~Makes sense, but we will be doing that in the future to ensure every icon remains the size it was meant to be. Right now the priority is to have the components finished exactly like they are on Figma.~~

Let's do it, but revert that default `#FFF` fill from the svgs. Pure white only applies to the red button children.

kibbeh/src/icons/solid-rocket.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="40" height="40" viewBox="0 0 40 40" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M36.0783 4.1874C36.0471 4.05459 35.9533 3.95303 35.8205 3.92178C31.5705 2.88271 21.7502 6.58584 16.4299 11.9061C15.3908 12.9452 14.5392 13.9686 13.8439 14.9608C12.2033 14.8124 10.5627 14.9374 9.16423 15.5468C5.21892 17.2811 4.07047 21.8124 3.75017 23.7577C3.67204 24.2187 4.05485 24.6171 4.51579 24.5702L10.8517 23.8749C10.8595 24.4843 10.8986 24.9686 10.9376 25.3046C10.9689 25.6327 11.1173 25.9374 11.3517 26.1718L13.8048 28.6249C14.0392 28.8593 14.3439 29.0077 14.672 29.039C15.008 29.078 15.4923 29.1171 16.094 29.1249L15.3986 35.453C15.3517 35.9139 15.7502 36.2968 16.2111 36.2187C18.1564 35.9061 22.6955 34.7577 24.4221 30.8124C25.0314 29.4139 25.1564 27.7811 25.0158 26.1483C26.008 25.453 27.0392 24.6014 28.0783 23.5625C33.4221 18.2499 37.1017 8.64834 36.0783 4.1874ZM23.3283 16.6796C22.0236 15.3749 22.0236 13.2655 23.3283 11.9608C24.6329 10.6562 26.7424 10.6562 28.0471 11.9608C29.3517 13.2655 29.3517 15.3749 28.0471 16.6796C26.7424 17.9843 24.6329 17.9843 23.3283 16.6796Z" fill="white"/>
```
benawad: can you remove the raw svgs

kibbeh/svgr.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  typescript: true,
+  replaceAttrValues: { "#000": "currentColor" },
```
benawad: why did you change from `#fff` to `#000`?

benawad:
The icon size normalization, breaks some other components
image
image

The current icons are "draft" icons so they are going to get switched out in the future. It's probably best to scrap the normalization for now and add it in when the "prod" icons are in.

kibbeh/src/icons/solid-rocket.svg
chancovsky: Do we want to remove all the raw svgs from the repo or just relocate them to an `assets` folder or something like that and just have the svg react components in `/icons`?

kibbeh/svgr.config.js
chancovsky: mistake, changed to `#fff`

chancovsky:
> The icon size normalization, breaks some other components
> image
> image
> 
> The current icons are "draft" icons so they are going to get switched out in the future. It's probably best to scrap the normalization for now and add it in when the "prod" icons are in.

The ScheduledTime is an alignment issue outside of svgs that I fixed. And I adjusted the dogenitro svg sizing a bit to be inline with the designs.  I can revert this all back to the original svgs I you think we aren't ready to use normalized icons.

benawad:
yeah if you wouldn't mind reverting, I want to do the normalization at a later point

benawad:
Also is it possible to cleanup the changes so it just shows the stuff you changed.

PR #668, opened 2021-03-09
Completed Tag UI Component

ermalsh:
Completed `Tag` UI Component for new design
Added Story w/ Controls


kibbeh/src/ui/Tag.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+interface TagProps { }
+
+export const Tag: React.FC<TagProps> = ({
+  children,
+}) => {
+  return (
+    <div
+      className={`bg-primary-600 select-none text-xs px-1 font-bold text-primary-100 justify-center items-center rounded`}
```
ajmnz: Background should be `gray-700` and `gray-600` on hover. Padding should be `px-2` on both the flag tag and the text tag.

kibbeh/src/ui/Tag.tsx
```diff
@@ -0,0 +1,16 @@
...
+      style={{ height: "22px" }}
+    >
+      {typeof children === 'string' ? `#` + children : children}
```
ajmnz: @benawad should we have the `#` by default?

ajmnz (changes requested):

kibbeh/src/ui/Tag.tsx
benawad: no remove the default `#`, there's no need to couple it together

PR #554, opened 2021-03-05
Hyper Hype Beast Refactoring 4 Pull request

arthurysong:
I refactored Beef.Users logic into the subdirectories created in Beef (access, mutations, and queries). @ityonemo , can you take a look at the pull Request and see if it makes sense?

PR #669, opened 2021-03-09
Hyper hype beast refactoring 5

ityonemo:
activates only after https://github.com/benawad/dogehouse/pull/554 is cleared


nadirabbas:
@ityonemo this needs to go into `staging` instead of `new-design`

benawad:
@ityonemo looks good, I'll merge after conflicts

PR #675, opened 2021-03-09
Turkish Translation for scheduledRooms and roomChat

originalls:
Fixed some Turkish translation issues and added translation for scheduledRooms and roomChat.


benawad:
we did a rebase that did some things to the history lol, would you mind pulling and opening a new pr?

PR #678, opened 2021-03-09
Removed [emotes soon] span as emotes has been implemented

BronzW:
I might be wrong here but emotes has been implemented and I guess we therefore don't need the [emotes soon] span.


ajmnz:
Emojis have been implemented, not emotes (like the ones Twitch has)

PR #679, opened 2021-03-09
Better German translation / Removed Emotes soon

StayWithMeSenpai:
german translations were werid imo so i tried to change em  a little bit while i was bored, i hope you like em


nadirabbas:
I don't think we need to remove `emotes soon`. We don't actually have the custom emotes this text is talking about.

benawad:
also can you open a new pull request after pulling from latest staging

PR #684, opened 2021-03-09
NotificationElement UI Component

jee1mr:
Based on https://github.com/benawad/dogehouse/issues/637

Hi, I created a base generic view for NotificationElement. I feel like a couple of things could be missing. I Will work on it based on your inputs. Also, I'm unable to resize the SVG to match the design. Tried everything, nothing worked. Would love some help with that.

Question:
1. Should I create the LiveNotification, NewRoomNotification, FollowNotification components using the base component or take a "type" prop with a default fallback and display the UI based on that?

Any other feedback is welcome.


ajmnz:
>     1. Should I create the LiveNotification, NewRoomNotification, FollowNotification components using the base component or take a "type" prop with a default fallback and display the UI based on that?

I'd go with types, as this will probably work with some kind of token. @benawad what do you think?

benawad:
you should be able to do `style={{ height: 20, width: 20 }}` on the svg components I setup.

I would create a new component for each one

jee1mr:
> you should be able to do `style={{ height: 20, width: 20 }}` on the svg components I setup.

@benawad That is just adding space around the icon. It's not increasing the icon size.

> I would create a new component for each one

I like this better too. Will go ahead with this.

jee1mr:
@benawad @ajmnz Implemented all the components. Please have a look.

kibbeh/src/ui/NotificationElement/FollowNotification.tsx
```diff
@@ -0,0 +1,55 @@
...
+}) => {
+  const icon = <SingleUser src={userAvatarSrc} size="sm" isOnline={isOnline} />;
+
+  const notificationMsg = (
+    <>
+      <a
+        className="font-bold"
+        {...(userProfileLink ? { href: userProfileLink } : {})}
+      >
+        {username}&nbsp;
```
ajmnz: Move all the whitespaces to the other strings, just in case we have to add some hover effects to the username in the future (such as underline).

ajmnz (changes requested):

kibbeh/src/ui/NotificationElement/GenericNotification.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export const GenericNotification: React.FC<GenericNotificationProps> = ({
+  notificationMsg,
+  time,
+  actionButton,
+  icon,
+}) => {
+  return (
+    <div className="flex items-center" style={{ width: "400px" }}>
+      <div className="mr-3" style={{ height: "39px", width: "39px" }}>
```
ajmnz: Use `w-6 h-6` instead of the inline `39px`. For `UserAvatar` use `sm` size.

ajmnz (changes requested):

kibbeh/src/ui/NotificationElement/FollowNotification.tsx
```diff
@@ -0,0 +1,55 @@
...
+    </>
+  );
+
+  const followButton = (
+    <Button
+      size="small"
+      color={followed ? "secondary" : "primary"}
+      style={{ width: "90px" }}
+    >
+      {followed ? "Followed" : "Follow back"}
```
ajmnz: Let's change `Followed` to `Following`. That way it doesn't look weird when a user that you are already following  follows you.

ajmnz (changes requested):

kibbeh/src/ui/NotificationElement/FollowNotification.tsx
jee1mr: Yes.

kibbeh/src/ui/NotificationElement/GenericNotification.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const GenericNotification: React.FC<GenericNotificationProps> = ({
+  notificationMsg,
+  time,
+  actionButton,
+  icon,
+}) => {
+  return (
+    <div className="flex items-center" style={{ width: "400px" }}>
```
ajmnz: Let them go full width, as they'll be inside a container.

ajmnz (changes requested):

kibbeh/src/ui/NotificationElement/GenericNotification.tsx
jee1mr: Done. Should I center align everything, or is this ok?

PR #685, opened 2021-03-09
Completed Tag UI Component

ermalsh:
Completed Tag UI Component for new design Added Story w/ Controls

### DIFF
--- a/kibbeh/src/ui/Tag.tsx
+++ b/kibbeh/src/ui/Tag.tsx
@@ -1,0 +1,16 @@
+import React from "react";
+
+interface TagProps {}
+
+export const Tag: React.FC<TagProps> = ({
+  children,
+}) => {
+  return (
+    <div
+      className={`bg-primary-600 select-none text-xs px-1 font-bold text-primary-100 justify-center items-center rounded`}
+      style={{ height: "22px" }}
+    >
+      {typeof children === 'string' ? `#` + children : children}
+    </div>
+  );
+};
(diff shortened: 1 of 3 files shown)


kibbeh/src/ui/Tag.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+interface TagProps { }
+
+export const Tag: React.FC<TagProps> = ({
+  children,
+}) => {
+  return (
+    <div
+      className={`bg-gray-700 hover:bg-gray-600 text-xs px-2 font-bold text-primary-100 justify-center items-center rounded`}
```
ajmnz: `bg-gray-*` is now `bg-primary-*`. For me it shows a transparent background in the preview

ajmnz: Add also `cursor-pointer`.

ajmnz (changes requested):

kibbeh/src/ui/Tag.tsx
ermalsh: @ajmnz sounds good! Should we update bg-gray-800 in MultipleUsers.tsx:15 and other places bg-gray is referenced? Those show up transparent in preview too.

ajmnz: ~~Sure, go ahead!~~ Wait, where do you see them being transparent? MultipleUsers shouldn't have any background I think

ermalsh: Done!

ajmnz: I think background doesn't do anything there, so we might be better off removing the class.

ajmnz (approved):

PR #693, opened 2021-03-10
SearchResult components

sahilmob:


ajmnz (changes requested):
~~I think we should remove the `hover:bg-primary-700` and make the room title, room hosts, username, etc. clickable, instead of the whole card.~~

ajmnz (approved):

Issue #698, opened 2021-03-11
Atomic Components Implementation v2


ajmnz:
<h1>Atomic Components v2</h1>

This is the follow-up for #637.

### Ready to be implemented

- [x] [Taken] RoomCard - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1355%3A831
    - Requires ScheduledTime, IconHeading & Tag ✅
- [x] [Taken] UserSummaryCard - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1356%3A771
    - Requires UserAvatar, UserBadge ✅
- [x] [Taken] GlobalSearch - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1356%3A949
    - Requires SearchBar, SearchHistory, SearchResult, SearchOverlay ✅
- [x] [Taken] SettingsDropdown - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1397%3A46
    - Requires SettingsIcon, BaseOverlay ✅
- [x] [Taken] NotificationsDropdown - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1397%3A143
    - Requires BaseOverlay, NotificationElement ✅
- [x] [Taken] MessagesDropdown - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1397%3A201
    - Requires BaseOverlay, MessageElement ✅
- [x] [Taken] MinimizedRoomCard - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1356%3A773
    - Requires Button, BoxedIcon ✅
- [x] [Taken] UpcomingRoomsCard - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1356%3A878
    - Requires BoxedIcon ✅

### Top level components

- [x] [Taken] FriendsOnline - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1397%3A474
    - Requires UserAvatar ✅
- [x] [Taken] Feed - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=1397%3A797
    - Requires Button, RoomCard ✅
[...]

PR #703, opened 2021-03-12
Add RoomCard component

lenilsondc:
This PR adds the `RoomCard` component as part of #698.


ajmnz:
~~Why isn't this deploying?~~

ajmnz (changes requested):
Let's set the scheduled time to `Month (abbr) Day` (Feb 20, Mar 14, etc.) when it isn't today. We'll show the time and details in a popup when the user clicks the card.

kibbeh/src/ui/RoomCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+function formatNumber(num: number): string {
+  return Math.abs(num) > 999
+    ? `${Math.sign(num) * Number((Math.abs(num) / 1000).toFixed(1))}K`
+    : `${Math.sign(num) * Math.abs(num)}`;
+}
+
+function useScheduleRerender(scheduledFor: Date) {
+  // same logic stolen from kofta, rerenders
+  // at the scheduleFor date
```
ajmnz, line 16: What if the host isn't online when the room starts? I'd probably leave a 15min margin before deleting/hiding the room instead of just going live directly, which might look weird and confusing. Let's wait for @benawad 's opinion.

lenilsondc: It doesn't hide anything right now, it only switches state from scheduled to live. As it's probably going to be used in `feed` it wouldn't make sense to `delete/hide` the card but make the room accessible perhaps? Or maybe people could access the room while waiting for the host, similar to YouTube lives where people can start chat while the host is starting.  But indeed it require more clarification on how it's going to work.

ajmnz: Then when that happens maybe change the time for `Soon`? But for me what would make sense is that the host had a `Go Live` button.

Being able to chat prior to the room starting would be interesting. Like being able to enter the room, having all the listeners listed and then having the start time in the speakers section.

benawad: I'll probably be changing when exactly it rerenders but that's an easy tweak we can do later.

The mockup is missing some actions for scheduled cards i.e. go live, edit, delete, add to calendar, # of people attending, cohosts and copy link

We can come back and edit `RoomCard` after we know what scheduled card looks like or maybe have it be a new component

lenilsondc: Oh I see, so this component is more like a `FeedCard` then. Anyhow I think we can resolve this for now. I'll push the other fixes.

PR #708, opened 2021-03-12
Global search component

sahilmob:
This PR includes:

-   Moving all search related components to Search folder
-   Creating GlobalSearch component


ajmnz:
The order has to go like this: `SearchBar` > click > `SearchOverlay` + `SearchHistory` > type > `SearchResult`.
Also, the overlay should match the `SearchBar` position (maybe by just removing the `SearchBar` in `SearchOverlay` and positioning it behind it?). Whatever feels best 👍🏼

ajmnz:
We also have an error in `RoomSearchResult`.

sahilmob:
> The order has to go like this: `SearchBar` > click > `SearchOverlay` + `SearchHistory` > type > `SearchResult`.
> Also, the overlay should match the `SearchBar` position (maybe by just removing the `SearchBar` in `SearchOverlay` and positioning it behind it?). Whatever feels best 👍🏼

As far as I understand, the overlay width has to be the same as the search bar and it has no padding top, is this right?
Also, what is the error in `RoomSearchResult`?

sahilmob:
search overlay

Is this correct? The way I approached that is by passing the focused state to the `SearchOverlay `component, and conditionally attaching `pb-5` class if the box is focused. Also, I removed other padding classes completely.

ajmnz:
> Is this correct? The way I approached that is by passing the focused state to the `SearchOverlay `component, and conditionally attaching `pb-5` class if the box is focused. Also, I removed other padding classes completely.

If you check the Figma file at page `Web App/Desktop` and hit the play button, you can see how the search overlay works.
The search bar stays in the same position, but the overlay goes `px-2 pt-2` from it. 

https://user-images.githubusercontent.com/57961822/110983259-d26b8a80-8369-11eb-8db5-f769f61a1827.mov

sahilmob:
@ajmnz Can you check the updated version please

PR #711, opened 2021-03-12
Implement MinimizedRoomCard

overlisted:
I'm not sure if that "inverse action flow" is how I was supposed to do the mute, deafen and leave buttons, but I tried.


ajmnz:
> I'm not sure if that "inverse action flow" is how I was supposed to do the mute, deafen and leave buttons, but I tried.

I'm not sure if you implemented that as I can't see it in the story, but don't worry about it now. We'll implement that in the future, as the icons aren't ready yet.

PR #718, opened 2021-03-12
Add relative link in readme

kennethnwc:
Add relative link for better navigation


benawad:
can you reopen for `new-design` branch since that's the default branch for readme

PR #744, opened 2021-03-12
Added Greek translation

nttzamos:


nadirabbas:
could you pull from `staging` instead of `new-design`. currently it will merge `new-design` into `staging`

Issue #705, opened 2021-03-12
Link to Discord stays in English even if selected another language


MattLawz:
**Describe the bug**
The discord link at the bottom stays in English instead of changing to the selected language.

**To Reproduce**
1. Change the website language at the bottom using the dropdown
2. See the error

**Expected behavior**
The Discord text should change to the selected language.
Example:
In dutch, discord should be "_onenigheid_"

**Video Showcase / Proof**
[Imgur Link](https://imgur.com/GcIa7aD)

**What device are you on?**
I'm on the Brave browser in Windows 10.
Brave Browser Version: `V1.21.74`

1kvsn:
Fix submitted as suggested in 'Expected Behavior'.

1kvsn:
Now, I think I might have jumped the gun on this. I apologise.

MattLawz:
> Now, I think I might have jumped the gun on this. I apologise.

All fine, I decided to also get into some code contribution so I did the translations myself and then pushed a PR. Just waiting for it to be merged,

ernestoittig:
Isn't this expected behavior? Discord is a brand name, they don't translate it on their own website.

DGKSK8LIFE:
> Isn't this expected behavior? Discord is a brand name, they don't translate it on their own website.

Yes, exactly.

ernestoittig:
Reminder that #742 that "fixed" this issue got merged

PR #745, opened 2021-03-12
Translated "Discord" to the selected language

MattLawz:
Fix for #705


nadirabbas:
could you pull from `staging` instead of `new-design`. currently it will merge `new-design` into `staging`

PR #747, opened 2021-03-12
updates Spanish translations

felipecespedes:
It translates new translation keys for Spanish


Giovanny-DS:
I can confirm this is actually a very good translation.

PR #785, opened 2021-03-12
Translated "Discord" to the selected language for most languages

MattLawz:
Fix for issue #745 and #705

**This is only for some languages, I didn't wanna touch about 5 because I wasn't familiar with them.**

### DIFF
--- a/kofta/public/locales/cs/translation.json
+++ b/kofta/public/locales/cs/translation.json
@@ -24,7 +24,7 @@
 	"footer": {
 		"_comment": "Main Footer UI Internationalization Strings",
 		"link_1": "Původ",
-		"link_2": "Discord",
+		"link_2": "svár",
 		"link_3": "Nahlásit chybu"
 	},
 	"pages": {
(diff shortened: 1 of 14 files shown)


nadirabbas:
Why would you translate Discord? It is a name...

prokopschield:
Is this a joke?

Discord is an app....

MattLawz:
> Why would you translate Discord? It is a name...

Well from what I know in some other countries, "discord" is in their language so they might not understand it because discord has the name to the location.

nadirabbas:
@prokopschield Discord is the name of an app, isn't it? Would you translate Dogehouse to Kuttaghar in urdu?

prokopschield:
> > Why would you translate Discord? It is a name...
> 
> Well from what I know in some other countries, "discord" is in their language so they might not understand it because discord has the name to the location.

Discord, as in, the app, is always Discord, in every language.

nadirabbas:
discord is an English word too, so that's why it has translations. but in the case of Discord is a proper noun.

nadirabbas:
so we dont need to translate Discord as far as I know.

MattLawz:
> @prokopschield Discord is the name of an app, isn't it? Would you translate Dogehouse to Kuttaghar in urdu?

I guess to

> so we dont need to translate Discord as far as I know.

Your right, you may close the PR.

PR #786, opened 2021-03-12
Implement Feed component

yvesnrb:
This PR implements the Feed component and associated story.

### DIFF
--- a/kibbeh/src/ui/Feed.tsx
+++ b/kibbeh/src/ui/Feed.tsx
@@ -16,23 +16,6 @@ export interface FeedProps {
   emptyPlaceholder: ReactNode;
 }
 
-export interface FeedHeaderProps {
-  title: string;
-  actionTitle: string;
-  onActionClicked: MouseEventHandler<HTMLButtonElement>;
-}
-
-export const FeedHeader: React.FC<FeedHeaderProps> = ({
-  title,
-  actionTitle,
-  onActionClicked,
-}) => (
-  <div className="justify-between items-end mb-5">
-    <h4 className="text-primary-100 ml-4">{title}</h4>
-    <Button onClick={onActionClicked}>{actionTitle}</Button>
-  </div>
-);
-
 export const Feed: React.FC<FeedProps> = ({
   title,
   actionTitle,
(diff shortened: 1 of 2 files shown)


ajmnz (changes requested):
I think user-subscribed scheduled rooms should be on top, but how would you like to go about this? @benawad

yvesnrb:
> I think user-subscribed scheduled rooms should be on top, but how would you like to go about this? @benawad

In that case, should I just add a 'subscribed' boolean to the props interface of Feeds to handle that logic and let whatever is referencing the Feed component handle it?

ajmnz:
> > I think user-subscribed scheduled rooms should be on top, but how would you like to go about this? @benawad
> 
> In that case, should I just add a 'subscribed' boolean to the props interface of Feeds to handle that logic and let whatever is referencing the Feed component handle it?

Kind of, but maybe we don't want to show more than x scheduled rooms or only rooms that are scheduled for today.

yvesnrb:
> > > I think user-subscribed scheduled rooms should be on top, but how would you like to go about this? @benawad
> > 
> > 
> > In that case, should I just add a 'subscribed' boolean to the props interface of Feeds to handle that logic and let whatever is referencing the Feed component handle it?
> 
> Kind of, but maybe we don't want to show more than x scheduled rooms or only rooms that are scheduled for today.

Got it. Just let me now how the feed should work @benawad and I'll make the necessary changes.

kibbeh/src/ui/Feed.tsx
```diff
@@ -0,0 +1,68 @@
...
+  rooms,
+  emptyPlaceholder,
+}) => {
+  const [liveRooms, setLiveRooms] = useState<RoomCardProps[]>([]);
+  const [scheduledRooms, setScheduledRooms] = useState<RoomCardProps[]>([]);
+
+  useEffect(() => {
+    setLiveRooms(rooms.filter((room) => isPast(room.scheduledFor)));
+    setScheduledRooms(rooms.filter((room) => !isPast(room.scheduledFor)));
+  }, [rooms]);
```
benawad: don't use useEffect and store rooms in state. Take rooms as a prop. Also don't worry about scheduled rooms for now.

yvesnrb:
I left the divider commented out for later and just mapped trough the rooms prop for now. Let me know if that is sufficient.

benawad (approved):

PR #804, opened 2021-03-13
feat: Electron wrapper global keybinds

amitojsingh366:
I've gone ahead and added global keybinds for the electron wrapper *(baklava)*. Global keybinds means that even if the app is in background the Keybinds will fire their events *(except PTT, for now it still uses the old detection)*. I had to change/add a few things on the front end *(kofta)* namely being able to send events to and from electron and the frontend *(made it so this only occurs if the webpage is opened in Electron)*. And also i did this right now and didn't wait for the new design as i didn't change any design and think that these changes can be easily carried over to the new design when it comes out.


baklava/src/constants.ts
```diff
@@ -1 +1,65 @@
...
+} from "electron";
+
 export const __prod__ = process.env.NODE_ENV === "production";
+export const REQUEST_TO_SPEAK_KEY = "@keybind/invite";
+export const INVITE_KEY = "@keybind/invite";
+export const MUTE_KEY = "@keybind/mute";
+export const CHAT_KEY = "@keybind/chat";
+export const PTT_KEY = "@keybind/ptt";
+export let CURRENT_REQUEST_TO_SPEAK_KEY = "Control+8";
+export let CURRENT_INVITE_KEY = "Control+7";
```
benawad: the bindings can change for keys

amitojsingh366: Yes im aware im using those variables for storing the current keybinds in cache *(The values set right now are just defaults)*. Is that bad practice? If yes please advise me on what to do as I feel this is one of the only ways to store data on the electron main process unless we store it in the `%appdata%`  folders or similar. Rather than that what my commit uses right now is how you were storing keybinds before, using `localStorage` on the Browser Window

amitojsingh366: The function right under these lines `RegisterKeybinds` sets listeners for any KeyBinds changing and removes the previous ones and sets the new ones as GlobalShortcuts

amitojsingh366:
Also @benawad i can take a shot at PTT even but give me a day or two as im a little busy rn. And im kinda new to git so im confused can i add more commits to this pull or how does it work?

baklava/src/constants.ts
benawad: nvm, I was confused about how this worked, but I see it now and it looks good

kofta/src/app/components/KeybindListener.tsx
```diff
@@ -1,22 +1,91 @@
...
+      });
+      setMute(!muted);
+    })
+    ipcRenderer.on(INVITE_KEY, async (event: any, args: any) => {
+      wsend({ op: "fetch_invite_list", d: { cursor: 0 } });
+      history.push("/invite");
+    })
+    ipcRenderer.on(CHAT_KEY, async (event: any, args: any) => {
+      toggleOpen();
+    })
```
benawad: do you need to unsubscribe on useEffect cleanup?

benawad:
Yeah you should be able to push up more commits

kofta/src/app/components/KeybindListener.tsx
amitojsingh366: Umm i dont think there's any need for it but electron does provide a function to remove the listener, so ig ill put that in cleanup. Will push a new commit soon with the PTT key even working

amitojsingh366:
@benawad I've gone ahead and added the PTT global keybind *(JANK WARNING)* and added the cleanups you requested

kofta/src/app/components/KeybindListener.tsx
```diff
@@ -1,22 +1,116 @@
...
+      ipcRenderer.removeListener(CHAT_KEY, CHAT_KEY_FUNC)
+    }
+  }, []);
+  return (
+    <GlobalHotKeys
+      allowChanges={true}
+      keyMap={keyMap}
+      handlers={useMemo(
+        () => ({
+          PTT: (e) => {
```
benawad, line 89: is this needed anymore?

baklava/src/util.ts
```diff
@@ -0,0 +1,145 @@
...
+                    PTT_STATUS[i] = false;
+                }
+            }
+            let PTT = PTT_STATUS.every((key_status) => key_status === true);
+            if (PTT != PTT_PREV_STATUS) {
+                PTT_PREV_STATUS = PTT;
+                mainWindow.webContents.send("PTT_STATUS_CHANGE", !PTT);
+            }
+        })
+        ioHook.start();
```
benawad: do you need unsubscribe logic if someone binds push to talk multiple times?

kofta/src/app/components/KeybindListener.tsx
amitojsingh366: yep it is cuz for some reason the `iohook` events dont fire when app is in foreground so i just let that stay there and it works perfectly

baklava/src/util.ts
amitojsingh366: ahh thats a good point. mistake on my part will fix it

amitojsingh366: lol wait i read your question wrong

amitojsingh366: No you dont need unsub logic cuz im comparing the pressed key with a variable. but your question made me realize i was adding multiple `keyup` and `keydown` events *(whenever the user changed PTT keybind new listeners would be added)*. Im fixing this rn

benawad:
nice work!

amitojsingh366:
> 
> 
> nice work!

😄

PR #817, opened 2021-03-13
Added package-lock.json to .gitignore

willdoescode:
No reason to have it on github


aidenybai:
Technically there is a point to have it on github, it ensures that the dependencies of the dependencies (and so on) are locked. However, this probably isn't super important

willdoescode:
Having the package-lock.json on github helps with addition farming as one new dependency or change can mean a lot of changes in package-lock.json

willdoescode:
removing it once and for all can fix this

nadirabbas:
but we still need the lockfile on GitHub. There can sometimes be breaking changes in newer versions.

benawad:
i want lockfile

PR #818, opened 2021-03-13
Add indonesia translation

chimemoo:


nadirabbas:

PR #820, opened 2021-03-13
Added Tranditional Chinese localization

selient:


nadirabbas:

selient:
@nadirabbas thanks, I opened another PR.

Issue #810, opened 2021-03-13
Links inserted into bio aren't parsed correctly.


mohammedgad:
**Describe the bug**
Links inserted into bio aren't parsed correctly check the screenshot.

**To Reproduce**
1. Go to https://dogehouse.tv/me
2. Click on Edit Profile
3. Enter two or three links in separate lines
4. Submit
5. Check your profile links now, all the links are combined into one link

**Expected behavior**
Links should be parsed separately and each link should be directed to the correct URL

**Screenshots**
![image](https://user-images.githubusercontent.com/8174938/111018651-b5af7100-83c2-11eb-9a61-85937856bc1e.png)

**What device are you on?**
Google Chrome on macOS

mohammedgad:
cc: @nadirabbas

nadirabbas:
yeah currently, the tokens are created by splitting by space, not by a line break

xiuxiu62:
Just wrote a parser to fix this

nadirabbas:
@justincremer I suggested some changes on your commit.

nadirabbas:
@justincremer this has been fixed #839

PR #836, opened 2021-03-13
Fix for #810

xiuxiu62:
This fixes a bug where hyperlinks in a user's profile were being conjoined if separated by a '\n' char.  The parser checks against regex whitespace instead.  The overhead is necessary, as the operation requires splitting chunks on different strings of whitespace and then rejoining, while maintaining the same string of whitespace - unachievable by using standard splits and joins (e.g. " " !== "    ").


nadirabbas:
fixed in #839 already.

PR #840, opened 2021-03-13
Further improve Norwegian language support

OscarWoHA:
This PR both fixes some of the translations for the Norwegian locale and adds support for disabling capitalization on buttons for languages where it just looks funky.


nadirabbas:
Please pull from `staging` instead of `new-design`

PR #846, opened 2021-03-13
fix: offline mention

### LINKED ISSUE #635: Disable clicking offline mentioned users in chat
**Describe the bug**
When you click a mention of a user that has disconnected it starts showing multiple errors on and on.

**To Reproduce**
1. Go to a room
2. Click on a mention of a disconnected user

**Expected behavior**
Disable clicking offline mentioned users in chat

**Screenshots**
Screenshot 2021-03-07 at 20 55 36

**What device are you on?**
Firefox Developer Edition, Apple MacBook Pro (Late 2018)

nadirabbas:
Fixes #635 

I could do a `where username or id` but this happens when we try that: 

image


kousa/lib/socket_handler.ex
```diff
@@ -769,9 +769,9 @@ defmodule Kousa.SocketHandler do
     end
   end
 
-  def f_handler("get_user_profile", %{"userId" => user_id}, %State{} = _state) do
-    user = Beef.Users.get_by_id(user_id)
-
+  def f_handler("get_user_profile", %{"userId" => id_or_username}, %State{} = _state) do
+    user = Beef.Users.get_by_username(id_or_username)
+    user = if is_nil(user), do: Beef.Users.get_by_id(id_or_username), else: user
```
benawad: Here's how you can check if the input is a `uuid` https://github.com/benawad/dogehouse/blob/staging/kousa/lib/routes/scheduled_room_routes.ex#L15

use that to know if you should lookup by id or username

PR #864, opened 2021-03-13
Fixed overflow of actionButton on baseOverlay due to no border-radius

yvesnrb:
The action button is poking out of it's container due to lack of border radius on the base overlay component.

Before:
before

After:
after


nadirabbas:
can't we do `overflow: hidden` on the parent container so the `button` doesn't overflow?

PR #875, opened 2021-03-13
Fix UserAvatar UI Component

Angshu31:
Fix some style choices

### DIFF
--- a/kibbeh/src/ui/UserAvatar/SingleUser.tsx
+++ b/kibbeh/src/ui/UserAvatar/SingleUser.tsx
@@ -45,7 +45,7 @@ export const SingleUser: React.FC<AvatarProps> = ({
 }) => {
   return (
     <div
-      className={`relative inline-block ${className}`}
+      className={`relative inline-block${className ? ` ${className}` : ""}`}
       style={{
         width: avatarSizeMap[size],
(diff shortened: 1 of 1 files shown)


ealmiladi:
You can just set a default for className, then you don't need the ternary in-line.
```
className = "",
```

Angshu31:
> You can just set a default for className, then you don't need the ternary in-line.
> 
> ```
> className = "",
> ```

Oh. What I did would get rid of the useless trailing space. Not that it matters much

kibbeh/src/ui/UserAvatar/SingleUser.tsx
```diff
@@ -45,20 +45,16 @@ export const SingleUser: React.FC<AvatarProps> = ({
 }) => {
   return (
     <div
-      className={`relative inline-block ${className}`}
+      className={`relative inline-block${className ? ` ${className}` : ""}`}
```
benawad: I'll take the trailing space over ternary

Angshu31:
I changed to trailing space

kibbeh/src/ui/UserAvatar/SingleUser.tsx
Angshu31: Done

PR #877, opened 2021-03-13
zh-CN: Fix some problems

itsjoeoui:


itsjoeoui:
"Microphone" was mistranslated. Fixed in this commit.

PR #879, opened 2021-03-13
Updated spanish translation

natanparrondo:


benawad:
switch to staging branch

PR #883, opened 2021-03-13
enhancement: A secondary modal/popup to confirm entering/leaving a room.

jacobmiller22:
added a confirmation modal when joining a room to prevent users from accidentally joining a room. I think this can also be refined more by making the modal give more details of the room the user is about to join. If that is something people think is useful I would be more than happy to implement. I am new to open source and I couldn't figure out how to get my formatter to not reformat the entire Home component. I only added about 6 lines of code give or take. Please leave any feedback or any dos/ do nots .

### DIFF
--- a/kofta/src/app/pages/Home.tsx
+++ b/kofta/src/app/pages/Home.tsx
@@ -1,4 +1,4 @@
-import React, {  useState } from "react";
+import React, { useState } from "react";
 import { Calendar } from "react-feather";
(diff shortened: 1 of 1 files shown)


benawad:
do this on the staging branch, this formatting should be fixed there. I only want a confirm if you are currently in a room and switch to another one (idk if you are already doing this, I haven't looked at the code yet)

PR #886, opened 2021-03-13
Added romanian translation

slendidev:


overlisted:
wtf

slendidev:
Sometimes, git is just not having a good day

slendidev:
The reason this has soooo many commits is because I was forced to do a git pull

PR #894, opened 2021-03-13
Added Romanian translation

slendidev:


nadirabbas:
please add it in the language dropdown too

slendidev:
Done

PR #907, opened 2021-03-13
Updated with better/worse translations

rapolasLT:
Maybe some translations are better. I don't know we have to compare them


ermalsh (approved):

nadirabbas:

PR #908, opened 2021-03-13
Added standard Arabic

ItsOmarKhater:
Please note that Arabic is an RTL language, so you'll need to consider this in the UI.


ermalsh:
Will need RTL support as stated, keep open until then.

nadirabbas:

PR #911, opened 2021-03-13
Implements logo component

yvesnrb:
Logo component is necessary for the Header.


ermalsh:
may need a story for it too if we're going to consider it a UI component

benawad:
we are not doing stories for the other icons, so I think this is fine for now

yvesnrb:
> may need a story for it too if we're going to consider it a UI component

I noticed the other svg components do not have stories of their own...
What's the course of action here @benawad ?

PR #916, opened 2021-03-13
Add Farsi(Persain)/Fa Translation

VahidTa:
Add Farsi translation. It needs RTL support.


overlisted:
we're definitely not gonna have an RTL layout

VahidTa:
So, I will close this one, and add new poll request and you can label it for enhancements when RTL support added.

PR #918, opened 2021-03-13
Add Farsi(persian) localization

VahidTa:
Add Farsi(Persian) translation.


nadirabbas:

PR #922, opened 2021-03-13
added translation for swahili language,

jameskaguo777:
fix for trailing comma in tr locales file,
Goodluck


nadirabbas:

PR #931, opened 2021-03-14
Fixing typo in Finnish translation

rzsxdev:
Changing "kopio linkki" to "kopioi linkki"


nadirabbas:

PR #934, opened 2021-03-14
Electron Wrapper (baklava) improvements in keybinds and builds

amitojsingh366:
- Changed build scripts 
- Replaced `electron-packager` with `electron-builder` to favor single file output *(`.msi`, `.dmg`, etc)*
- Replaced the previous icons with the new icons
- Added author and repo info for linux builds *(`electron-builder` needs it only for linux builds)*
- Added `iohook` stop sequence to prevent issues *(quitting dogehouse build on mac caused warnings)* 
- Added proper keystroke detection as i noticed my previous merged pull request *([#804](https://github.com/benawad/dogehouse/pull/804))* wasn't working as expected *(issue turned out to be that `iohook` used linux keycodes and not JS)*

> **NOTE:** I have tested the windows and macOS versions and they run fine but no testing has been done on the linux build


amitojsingh366:
@benawad need a review...

PR #935, opened 2021-03-14
Fix hebrew translations & microcopys 

alon-abadi:


nadirabbas:

PR #936, opened 2021-03-14
Swahili language translation added

jameskaguo777:
Swahili language translation added

Good luck!


nadirabbas:
please also add it in the homepage language dropdown

jameskaguo777:
> please also add it in the homepage language dropdown

got it

PR #937, opened 2021-03-14
Fixes #443: Show spinner while fetching follower list

wes337:
Adds a loading state value to the follower atoms, and will show the spinner in the list while they are fetching.

Also noticing there are some pretty big diffs because of the white space formatting done by Prettier 😬, is this okay? 

You can test this by throttling your network speed, then opening clicking the "X followers" or "X following" in the user's profile modal. You should see a spinner instead of "no users found".


nadirabbas:

PR #942, opened 2021-03-14
Fixes #443: Show spinner while fetching follower list

wes337:
Adds a loading state value to the follower atoms, and will show the spinner in the list while they are fetching.

1. Throttle your network speed,
2. Click the "X followers" or "X following" in the user's profile modal. 
3. You should see a spinner instead of "no users found".


ermalsh (approved):
lgtm

kofta/src/app/pages/FollowListPage.tsx
```diff
@@ -29,6 +30,17 @@ export const FollowListPage: React.FC<FollowListPageProps> = () => {
...
+	useEffect(() => {
+    const fn = isFollowing ? setFollowingMap : setFollowerMap;
+    fn((m) => ({
+      ...m,
+      [userId]: {
+        ...m[userId],
+        loading: true,
+      },
+    }));
+	}, [isFollowing, setFollowerMap, setFollowingMap, userId]);
```
benawad: the ws could return the data before useEffect is called and it would infinitely load, set loading to true when the wsend is called

wes337: @benawad I'm a big fan, thank you so much for taking the time to review!

I used the same setFollowing/FollowerMap atom function inside the user profile, which is where the follow list is fetched from. Maybe this function could be abstracted into some other file, but I'm trying to keep the git diff small here.

PR #943, opened 2021-03-14
Add Romanian translation and run sync

mihaiyoo6:
Add Romanian translation and run sync


nadirabbas:

PR #948, opened 2021-03-14
Added greek translations 

toniker:
Added greek translations to json file
Added greek option in language selection list


manaskhandelwal:
@toniker [Greek translation](https://github.com/benawad/dogehouse/blob/new-design/kofta/public/locales/el-GR/translation.json) is already added.

ermalsh:
el-GR is already added.

PR #951, opened 2021-03-14
Fix RoomSearchResult story props mismatch

demfabris:
RoomSearchResult story props does not match the UI component's props, making it crash inside story book


tejasag:
lgtm

ermalsh (approved):

Issue #947, opened 2021-03-14
Ukrainian language support 


ArtemKutafin:
**Is your feature request related to a problem? Please describe.**
Ukrainian language support is missing

**Describe the solution you'd like**
I will add a json with translations

PR #954, opened 2021-03-14
feat: Add Ukrainian language #947

ArtemKutafin:
This PR adds Ukrainian language translation support. It also adds the language key to the `LanguageSelector.ts`.


overlisted:
isn't the key called `uk`?

overlisted:
the directory's name is still `ukr`

ermalsh (approved):

PR #958, opened 2021-03-14
Update German Translation

Khosraw:
Dutch was used and Discord is supposed to be used as a noun.


nadirabbas:
Please pull from `staging` instead of `new-design`.

PR #959, opened 2021-03-14
Updated German Translation

Khosraw:
It no longer says starting a room with a user. Now it correlates with the English translation of starting a room with "them".


MichaelTrenker:
Your translation made it not better (worse)!

nadirabbas:
Please pull from `staging` instead of `new-design`.

PR #976, opened 2021-03-14
CodeBlocks & text area instaid of input field

theCow61:
#961 @kleo-xhindoli @dhenson02 @nadirabbas helped me with this.
Does code blocks by inputting 3 back ticks like \```  then putting code then another \``` to close it. Input fields are now text areas, to get a new line you can do Shift+Enter like in telegram or whatsapp and enter for enter.

image

This would also make it easier for group teaching and learning of "dogelang" in dogehouse.


benawad:
0. I don't want a textarea
1. Use `staging` branch
2. Idk if I ever want code blocks, chunks of code should probably be a link to a gist instead of us trying to handle it, but I will think about this later
3. I am open to single \`  for code to create `these things`
4. Create separate PR for spacing fix first so its easier to see diffs
5. Don't check for backticks when the chat renders that will hurt performance. Instead check when the message is first created and create a new token type

PR #986, opened 2021-03-15
Add room time elapsed with moon travel status

### LINKED ISSUE #415: Reward for Rooms.
At Dogehouse, It is normal for public rooms to last for more than 16 to 18 hours, one thing we can do to provide feedback to speaker and listener is to reward them.

Specs :
1- A timer component will be rendered somewhere on the room(Team will decide), to let people know how much time is elapsed since its creation, changing the name of the room shouldn't reset the timer and it will be destroyed once the room is destroyed. 

2- A hidden custom party popper (picture / gif) will be triggered if the room has elapsed more than 24 hours or other time-limit that the team decides.

This will provide a good feedback to users just like game rewards system.
If @benawad  likes the idea, I can get it done.

NEO97online:
This PR adds an indication of how long each room has been active for. Rooms have been taking off lately and it would be great to see how close to the moon each room is. So along with the time indicator, this PR adds icons and a "rocket status" as a bonus for keeping a room active for a while.

This should mostly cover #415, but we could potentially take that issue further.

No extra logic is added on the backend for this, its just using the`inserted_at` timestamp.

image
image

### DIFF
--- a/kofta/src/app/utils/timeElapsed.ts
+++ b/kofta/src/app/utils/timeElapsed.ts
@@ -1,21 +1,19 @@
 import formatDistanceToNowStrict from 'date-fns/formatDistanceToNowStrict';
 import differenceInMinutes from 'date-fns/differenceInMinutes';
-import { RefObject, useEffect, useState } from 'react';
-import { Room } from '../types';
+import { useEffect, useState } from 'react';
 
-export const useTimeElapsed = (roomRef: RefObject<Room>) => {
+export const useTimeElapsed = (startDate: Date | null) => {
     const [timeElapsed, setTimeElapsed] = useState("");
     const [rocketIcon, setRocketIcon] = useState("");
     const [rocketStatus, setRocketStatus] = useState("");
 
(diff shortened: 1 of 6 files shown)


kofta/src/app/utils/timeElapsed.ts
```diff
@@ -0,0 +1,53 @@
+import formatDistanceToNowStrict from 'date-fns/formatDistanceToNowStrict';
+import differenceInMinutes from 'date-fns/differenceInMinutes';
+import { RefObject, useEffect, useState } from 'react';
+import { Room } from '../types';
+
+export const useTimeElapsed = (roomRef: RefObject<Room>) => {
```
benawad: instead of a ref, take a date parameter

benawad (changes requested):

kofta/src/app/utils/timeElapsed.ts
NEO97online: Good point, just cleaned up the refs.

Issue #979, opened 2021-03-14
Block certain words from being used in username field


alon-abadi:
**Is your feature request related to a problem? Please describe.**
It is possible to use abusive and harmful words in the username field. This field is actually used in the chat whenever someone tags that user, so for example if someone used the n-word as their username, when users tag that user, their messages actually uses that word. (true story)

**Describe the solution you'd like**
Filter/block certain words from being used in the username field.

**Describe alternatives you've considered**
In my opinion, blocking is necessary. But as an additional fix, it is also possible to use the user's display name in chat tags (makes sense) instead of their username. 

**Additional context**

juliankrieger:
I recommend using https://github.com/LDNOOBW/List-of-Dirty-Naughty-Obscene-and-Otherwise-Bad-Words

PR #993, opened 2021-03-15
Block NSFW words from being usernames

yummyweb:
This PR resolves/implements issue #979. The wordlist used comes from [here](https://github.com/LDNOOBW/List-of-Dirty-Naughty-Obscene-and-Otherwise-Bad-Words).


proudmuslim-dev (approved):

kofta/src/app/utils/getBadWords.ts
```diff
@@ -0,0 +1,15 @@
...
+
+// This will get the bad words from the txt file and return the array
+export const getBadWords = (): Array<string> => {
+    let returnArr: string = ''
+
+    fs.readFile('../../list_of_bad_words.txt', 'utf8', function (err,data) {
+        if (err) {
+          return console.log(err);
+        }
+        returnArr = data
```
NEO97online, line 11: `fs.readFile` is asynchronous. The callback function executes *after* the enclosing function returns, so `returnArr` will still be empty when `getBadWords` returns.

That said, this is a front-end web app; `fs` is a Node.js module. Instead of fs, just store an array of all words in a json file and import it that way.

yummyweb: ok

NEO97online:
If we're going to add this it should really be a backend validator, not on frontend. I also have concerns about this word list, it seems far too broad. For example if someone's name is `Hassan` it would be blocked for containing `ass`. I think a much smaller list of only the most offensive words would suffice.

kofta/src/app/utils/getBadWords.ts
yummyweb: tho i think fs would still work.

NEO97online: `fs` does not work on frontend, there is no filesystem.

yummyweb:
It is a valid point, but it won'twork like that. It would check if the word, 'ass' is 'ass' not if Hassan contains 'ass'. It would directly match it.

NEO97online:
If you only check for the entire word, then circumventing the filter is as simple as adding an extra character at the end.

yummyweb:
But that wouldn't exactly be NSFW is the conventional sense. So umm, I can either filter the words and check if the entire word contains multiple NSFW words or i could do something else.

l-mbert:
If `fs` is only used in Server-Side Functions it should work. 
The Question remains if this is really something for the Frontend. 

The Problem with Wordfilters is always: do we want False-positives or False-negatives. 
"Badass007" wouldn't be so bad. "YoureAnAss1337" is a different story. 

My Idea would be:
- If a User submits a Username which is not at all matched with our Filter, let him/her use it right away.
- If it's matched with the Filter, ask him/her to try again, if they insist let them use it but put them on a List to "review" by Moderators. (Which would be more Work for the Moderators.)
- Hope that the Community report bad Usernames.

Maybe you Guys have another great Ideas!

kousa/.envrc
```diff
@@ -1,9 +1,14 @@
...
+export DATABASE_URL=postgres://postgres:postgres@localhost/kousa_repo2
+export BEN_GITHUB_ID=7872329
+export RABBITMQ_URL=amqp://user:password@yourinternalip:5672
+export ACCESS_TOKEN_SECRET=
+export REFRESH_TOKEN_SECRET=
+export GITHUB_CLIENT_ID=efb151849cf1c786414e
+export TWITTER_API_KEY=
+export TWITTER_SECRET_KEY=
+export TWITTER_BEARER_TOKEN=
+export GITHUB_CLIENT_SECRET=ff64d1ff2ee89fcbf8f389e0f2b9de5ae1de849a
```
amitojsingh366, line 10: I don't think mentioning your `client-id`  and `client-secret` would be a good idea in here...
Also i don't think including this file itself would be a good idea.

kofta/src/app/utils/getBadWords.ts
amitojsingh366: Also @yummyweb looking at how dogehouse is structured i think all of the logic should be on the back end, people can circumvent this easily

amitojsingh366 (changes requested):
All of this logic should be on the back end

benawad:
yeah this needs to happen server side

kofta/src/app/utils/getBadWords.ts
juliankrieger: I agree

juliankrieger:
Not only that, it needs to be structured in a way that the file is not reach every time a call is made. I'd just store them in an array.

PR #1004, opened 2021-03-15
Improving the electron wrapper 

amitojsingh366:
- Fixed issue where environmental variables weren't working in the built app
- Fixed issue where the app would mess with the OS if multiple instances of it were opened *(permitted only one instance to open at a time)*
- Changes to `HOW_TO_RUN.md` to make it easier for people to be able to run this locally 
- External link handling implemented *(external links will now open in default browser)*


TheOtterlord (commented):
~~This is the wrong way round isn't it?~~ nvm, read it wrong XD

baklava/src/electron.ts
```diff
@@ -28,7 +28,7 @@ function createWindow() {
     mainWindow.webContents.openDevTools();
   }
   mainWindow.loadURL(
-    __prod__ ? `https://dogehouse.tv/` : "https://dogehouse.tv/"
+    __prod__ ? `https://dogehouse.tv/` : "http://localhost:3000/"
```
TheOtterlord: Here

amitojsingh366:
How did i manage to spell electron wrong twice....

baklava/src/electron.ts
amitojsingh366: no that's correct

amitojsingh366: if `__prod__` is `true` choose the first URL otherwise choose the 2nd url

PR #1005, opened 2021-03-15
Change in Dutch Translation

daankent:
Changed "volgt" to "volgend".
I think this is a better translation. This is also how Instagram translates it.


ermalsh:
Please create PR into staging and not new-design

PR #1024, opened 2021-03-15
Baklava: Discord-Like System Tray & Auto-Hide Default Menu

TheOtterlord:
Updates to the Electron app.

- Discord-like system tray (click to focus, right click to see context menu)
- Auto-hide the default electron menu (can access with `alt` if you need to. seems useless though, until we make our own)


TheOtterlord:
Sorry, wrong target

PR #1035, opened 2021-03-15
Update Dutch translation

daankent:
Changed 'volgt' to 'volgend'.
I think this is a better translation.
This is also how Instagram translates is.


ermalsh:
Was updated in #1055

PR #1036, opened 2021-03-15
Add translation keys for rocket->moon status

NEO97online:
Rocket->moon statuses were hardcoded in English, this PR moves them to translation keys.


ermalsh:
Please run 'npm run i18' so that the translations sync for the other locales.

ermalsh:
A lot of new languages were merged in, you'll have to fix those conflict.
Fetch latest staging from upstream and sync your branch up so it's equal with benawad:staging then run `npm run i18`.

PR #1038, opened 2021-03-15
Electron Splashscreen

yummyweb:
A working splash screen which currently just shows the DogeHouse logo on startup.


.editorconfig
```diff
@@ -6,6 +6,6 @@ root = true
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
-indent_style = tab
+indent_style = space
```
theotterlord, line 9: Probably best to keep with the previous configuration

baklava/src/electron.ts
```diff
@@ -11,12 +11,18 @@ import {
 import iohook from "iohook";
 import { RegisterKeybinds } from "./util";
 import { ALLOWED_HOSTS } from "./constants";
+import url from 'url'
+import path from 'path'
+
 let mainWindow: BrowserWindow;
 let tray: Tray;
-export const __prod__ = app.isPackaged;
+export const __prod__ = true; // app.isPackaged
```
theotterlord, line 19: This should be `app.isPackaged` so that developers can test locally

baklava/src/electron.ts
```diff
@@ -25,7 +31,18 @@ function createWindow() {
     webPreferences: {
       nodeIntegration: true,
     },
+    show: false
   });
+
+  // set splash screen
+  splash = new BrowserWindow({width: 810, height: 610, frame: false, transparent: true, alwaysOnTop: true});
+  console.log(`${__dirname}../splash-screen.html`)
```
theotterlord, line 39: Remove development logs?
```suggestion
```

TheOtterlord (changes requested):
Not a maintainer, but I've spotted a few things that could be looked at. 
Also, this seems like a big PR for a small change. 
Have you made an incorrect merge to your branch?

.editorconfig
yummyweb: i dint do this

baklava/src/electron.ts
yummyweb: yes, that was for my development

.editorconfig
theotterlord: Probably automatic IDE weirdness, but it will still make the change as it is in your branch

baklava/src/electron.ts
theotterlord: I guessed as much, I do the same thing sometimes, but it needs to be reverted before commiting to keep the same functionality for others

baklava/package.json
```diff
@@ -59,6 +59,7 @@
     }
   },
   "dependencies": {
+    "@trodi/electron-splashscreen": "^1.0.0",
```
theotterlord: I'm not seeing this used anywhere.
Unless I'm mistaken, it is an unused module and can be removed?

TheOtterlord (changes requested):

yummyweb:
This PR has many errors, so creating new pr.

PR #1041, opened 2021-03-15
Electron Splash Screen

yummyweb:
Working splash screen for electron client


TheOtterlord (changes requested):
tbh I think you may need to reclone/refork the repository as there are still over 100 files that have changes, most of which don't have anything to do with the Electron app. I think it's probably due to some merge somewhere or just an outdated fork

benawad:
do your work on latest staging to avoid all these commits

PR #1048, opened 2021-03-15
Corrected spelling mistakes & matched letter cases

vassbo:


ermalsh:
Source branch needs to be staging instead of new-design. @vassbo

PR #1049, opened 2021-03-15
working splashscreen

yummyweb:
A working splash screen for electron client.


yummyweb:
PS: I changed the __prod__ to true so I cannot set it to app.isPackaged again.

PR #1052, opened 2021-03-15
Resolves #572 - Hide/Show Listeners

### LINKED ISSUE #572: Minimize Listeners
Making it possible to minimize the listeners as listeners grow in number.

wes337:
If a room has more than 50 listeners, we will show a button which can be used to hide/show the listeners. Still needs a better design. Resolves #572

### DIFF
--- a/kofta/src/app/pages/RoomPage.tsx
+++ b/kofta/src/app/pages/RoomPage.tsx
@@ -48,6 +48,7 @@ export const RoomPage: React.FC<RoomPageProps> = () => {
   const fullscreenChatOpen = useShouldFullscreenChat();
   const [showCreateRoomModal, setShowCreateRoomModal] = useState(false);
   const [listenersPage, setListenersPage] = useState(1);
+  const pageSize = 25;
   const { t } = useTypeSafeTranslation();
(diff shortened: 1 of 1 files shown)


11ma:
> If a room has more than 50 listeners, we will show a button which can be used to hide/show the listeners. Still needs a better design. Resolves #572 

isn't 50 too much? won't that just make a room page unnecessary long to scroll down?

benawad:
do you have a screen shot of what it looks like?

wes337:
> 
> 
> > If a room has more than 50 listeners, we will show a button which can be used to hide/show the listeners. Still needs a better design. Resolves #572
> 
> isn't 50 too much? won't that just make a room page unnecessary long to scroll down?

It could be, what number do you think is a good amount before we allow the user to hide?

> 
> 
> do you have a screen shot of what it looks like?

Listeners open:
open

Listeners hidden:
closed

11ma:
noice, thanks for the screenshots, how does 25 sound?

benawad:
I think you should have it still show some listeners, but a show more at the bottom

PR #1060, opened 2021-03-15
Updated sound effects

catpieyum:
Added self made sound effects that fit better together and fixed already existing one (roomChatMention)


NEO97online:
These sound great imo

catpieyum:
If you need any more sound effects for stuff, let me know :)

ut 16. 3. 2021 o 1:45 Ben Awad ***@***.***> napísal(a):

Issue #972, opened 2021-03-14
External links open in the same tab


supremeicecreme:
When clicking on a link (specifically on the home page, I haven't got further than that yet!), they open in the same tab.  This is frustrating as you can lose where you are on the site previously.

Simple solution could be to add the `target="_blank"` tag to all external links so they pop open a new tab.  I think this is the best way, though I'm not sure.

I don't think there are any better alternatives to my suggestion, though research might prove otherwise.

Gif screen capture is attached so you know what I'm on about!
![LbEByNRVWq](https://user-images.githubusercontent.com/22503550/111084002-5c515a00-8508-11eb-9234-eccc003fd844.gif)

Screen capture where `target="_blank"` is added to the `Open Source` link
![j4Uz12ul5s](https://user-images.githubusercontent.com/22503550/111084083-9ae71480-8508-11eb-8c99-316fd82b105e.gif)

supremeicecreme:
I thought it best to open an issue with this and leave it to the people who know the app best! I've just come across the project and so I'm unfamiliar with the codebase and how it all works.

ermalsh:
Makes sense, I think it's a good idea

PR #1071, opened 2021-03-16
fix: #972 External links open in the same tab

MadanBhandari:


nadirabbas:
please also add `rel="noreferrer"` it is recommended by `eslint`.

PR #1086, opened 2021-03-16
Added en-doge and ran i18

HonbraDev:
Much bork


HonbraDev:
image

nadirabbas:
do you know why there are huge changes in other translation files too?

HonbraDev:
@nadirabbas yes, I ran `npm run i18`. If it's a problem, I'll close this one and make a new PR without running i18.

kofta/public/locales/en-doge/translation.json
```diff
@@ -0,0 +1,224 @@
+{
+  "_comment": "if you change this file, do: npm run i18",
```
HonbraDev, line 2: here

nadirabbas:
what does `i18n` do? doesn't it just add translation key types? You only run `i18n` when you manually add a new translation key. (as far as I know)

PR #1092, opened 2021-03-16
Only deploy on changes to `kibbeh`

TheOtterlord:
Deploy only when changes are made to `kibbeh`.


benawad:
I have that manually off right now, but it actually should go whenever `kousa` changes

PR #1102, opened 2021-03-16
Electron wrapper - disable dev tools in production

robertlit:
Disabled dev tools shortcut (Control + Shift + I) when in production.


benawad:
why? VSCode keeps it on

TheOtterlord:
Yeah. Plus, people in the know like to peek at the console log and sometimes the network tab if things go wrong.

BronzW:
Yeah, especially for an open source project DevTools would be pretty nice to have for anybody who would like to look in under the hood on the real website. It's not a security issue.

PR #1117, opened 2021-03-16
search icon redirect to searchUserPage.tsx

logark12:


benawad:
this is hidden on purpose atm

PR #1126, opened 2021-03-16
fixed a bug that caused chat text not to be translated

AuroPick:
image

Is there a quick way to add keys to translation files? Because I did it all one by one


ermalsh:
Fix conflicts :)

PR #1143, opened 2021-03-17
unique username urls

### LINKED ISSUE #1079: Distinct urls for users.
I am frustrated when I go to share mine or someone else's profile and the link is just dogehouse.tv/user or dogehouse.tv/me

I think it would be nice to have a url like `dogehouse.tv/@username` so that you can look at other peoples profiles as well as share your own that way.

**Describe alternatives you've considered**
I have also considered just having uuids for the url in place of usernames but I feel like that would be hard to use.

nadirabbas:
fixes #1079

dogehouse.tv/user/username


willdoescode:
Thanks Nadir!

PR #1146, opened 2021-03-17
Removing alwaysOnTop property of splash screen

yummyweb:
This pr is based on a feature request by @amitojsingh366 where he said that the splash screen should not always be on top. So this pr does exactly that.

Tested on MacOS Mojave 10.14


nadirabbas:
What's the rationale here? why shouldn't it be on the top tho?

amitojsingh366:
@nadirabbas the always on top property makes it so it even comes over other apps. So if you're doing something and open dogehouse side by side you have like 2 seconds where you aern't able to perform any actions as dogehouse is on top of other apps

amitojsingh366:
But i don't see why it requires 4000 lines of code. But i don't understand git much yet either.

nadirabbas:
please reset the `pacakage-lock.json` to `ben/staging`

yummyweb:
ok

yummyweb:
@nadirabbas done. Please merge if the pr is fine.

nadirabbas:
there are still changes in the lockfile. please reset

PR #1163, opened 2021-03-17
ReactNative App starter

Etienne-io:
This is a beginning for the ReactNative application.

- Splash screen + App Icon (I used what i had so we will have to update them later)
- Update app name
- Add github signin logic (target staging server for the moment). Opens a InApp webpage and redirect to the application via Deeplink
- Adding 'new-design' theme color and font
- Ionicons for temporary Icon
- ReactNavigation to handle screen navigation
- Default page (Home, Schedule, Following, Landing, Profile)


overlisted:
this is quite a big diff

Etienne-io:
There is not a lot of code in it. Most of the files changed are images, icons and fonts

milansusnjar:
There is React Native 0.64.0, it would be better to use it now to avoid issues with an upgrade later.

PR #1168, opened 2021-03-17
Adding of the Limburgish dialect/language

Laurenshup:
Limburg is region in the Netherlands.
Limburgish (or Limburgan/Limburger) is the dialect/language spoken there. The language has some similarities between Dutch and German.


ermalsh:
Changes were made to the LanguageSelector involving native language strings, please fix that conflict and then this should be good to go!

Issue #1097, opened 2021-03-16
option to handover room admin role


nadirabbas:
currently its random, sometimes unwanted people become owners

jamesql:
I'd be interested in working on this when I get home in a few hours if no one does it by then.

// had people complain about this to me a few hours ago

nadirabbas:
@benawad

jamesql:
should the previous creator be a mod after they have transferred ownership?

nadirabbas:
yeah, that makes sense @jamesql

jamesql:
i should have it done by the end of the day

benawad:
sounds good, go for it!

goldyydev:
Going to close this as it appears that it got added already.

PR #1178, opened 2021-03-18
add option to handover room admin

jamesql:
per #1097

- Adds option to change room owner or 'creator'
- Fixes typo li/translations.json -> li/translation.json


kousa/lib/kousa/room.ex
```diff
@@ -165,6 +165,23 @@ defmodule Kousa.Room do
       )
     end
   end
+	
+  def change_room_creator(new_creator_id, current_room_id \\ nil) do
```
nadirabbas: - reset `package-lock.json`
- also need to change room creator in the DB, `leave_room` sets the new creator for example https://github.com/benawad/dogehouse/blob/01dd6b5c18920a3458824fae1db48a4245f3aa8e/kousa/lib/beef/mutations/rooms.ex#L115
- backend validation: 
    - check if the user who initiated the request is the room creator of the room which they're in
    - check if the user_to_be_new_creator is in the same room as the room creator

nadirabbas (changes requested):

kousa/lib/kousa/room.ex
jamesql: no problem

nadirabbas: @jamesql I appreciate it

jamesql:
im going to test the new commits now give me about 10 minutes

kousa/lib/kousa/room.ex
jamesql: @nadirabbas take a look at it now, let me know if anything else should be changed

jamesql:
okay package-lock should be good now haha

PR #1193, opened 2021-03-18
updated hebrew translation

elad-weiss:


goldyydev:
@elad-weiss why did you close this pr? the translation itself is valid

elad-weiss:
Found another key that was not translater and didn't want to create two different pull requests

PR #1208, opened 2021-03-18
fixed slovak translation

AuroPick:
fixed slovak translation that caused i18 sync error


daloes:
Oops sorry my bad, I was editing on mobile 😅

PR #1210, opened 2021-03-18
added translation to user profile and follow list page

AuroPick:
added translation to user profile and follow list page, now follow button translating


nadirabbas:
why did you have to add the same keys twice?

AuroPick:
I did because they are two different page. Shouldn’t i

ermalsh:
Please fix conflicts. Also, if it's a common repeating translation key, just add it to commons instead of it's own category.

ermalsh:
One last conflict to fix and should be good to go. @AuroPick

Issue #1214, opened 2021-03-18
Error when running npm install in `kibbeh`


Mayowa-Ojo:
**Describe the bug**
Tried to install dependencies in `kibbeh` but got a Not Found error.

```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@dogehouse%2fkebab - Not found
npm ERR! 404 
npm ERR! 404  '@dogehouse/kebab@^0.1.2' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 It was specified as a dependency of 'kibbeh'
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

**To Reproduce**
1. Go to 'kibbeh'
2. Run  'npm install'

**Expected behavior**
Should install dependencies without errors.

**What device are you on?**
OS: Linux
Node version: v14.16.0
npm verson: v6.14.11

**Additional context**

ThatOneHidde:
😅

nadirabbas:
It's called `@dogehouse/client` now, maybe update the `package.json`

Mayowa-Ojo:
Yh, thanks. I think we should include a README that explains how to setup the next.js frontend  in `kibbeh`. I also had to switch to yarn because I learned the project now uses yarn.

martines3000:
Looks like I accidentally made the same commit twice... Still learning how the "npm run commit" command works. You can ignore the first commit. :sweat_smile:

goldyydev:
#1233

PR #1228, opened 2021-03-19
fix(@dogehouse/client dependencies in pages)

RocKhalil:
related to PR #1227 and bug #1214


goldyydev:
The package shouldnt've been renamed to `@dogehouse/client` in the first place. please refer to #1233 for more info

RocKhalil:
Makes sense, fixed it because it was merged and it wasn't working anymore

PR #1230, opened 2021-03-19
Build DogeHouse exe

yummyweb:
Added a script for building DogeHouse exe from electron-builder


baklava/package.json
```diff
@@ -1,57 +1,70 @@
 {
-  "name": "dogehouse",
-  "version": "1.0.14",
-  "description": "Taking voice conversations to the moon 🚀",
-  "main": "./dist/electron.js",
-  "scripts": {
-    "compile": "tsc",
-    "watch": "tsc -w",
-    "dev": "npm run compile && node ./dist/dev.js",
-    "start": "npm run compile && electron ./dist/electron.js",
```
theotterlord: no, not my dev scripts!

I think your fork may be outdated

baklava/package.json
```diff
@@ -1,57 +1,70 @@
...
+    "name": "dogehouse",
+    "version": "1.0.0",
+    "description": "Taking voice conversations to the moon 🚀",
+    "main": "./dist/electron.js",
+    "scripts": {
+        "compile": "tsc",
+        "watch": "tsc -w",
+        "dev": "npm run compile && electron ./dist/electron.js",
+        "build": "npm run compile && electron-builder",
+        "build:all": "npm run compile && electron-builder -mwl --win msi",
```
theotterlord: You're still requesting an msi build for windows

baklava/package.json
```diff
@@ -1,57 +1,70 @@
...
+        "build:mac": "npm run compile && electron-builder --mac",
+        "build:win": "npm run compile && electron-builder --win msi",
+        "build:exe": "electron-builder -p never --win",
```
theotterlord: If building an msi does not work, probably best to remove it

baklava/package.json
```diff
@@ -1,57 +1,70 @@
...
-      "category": "public.app-category.social-networking",
-      "entitlements": "settings/entitlements.mac.plist",
-      "extendInfo": {
-        "NSMicrophoneUsageDescription": "Mic access for speeach input in voice."
-      },
-      "target": [
-        "dmg"
-      ]
+    "name": "dogehouse",
+    "version": "1.0.0",
```
theotterlord: The version number should not go backwards. Again, I think you have an outdated fork

TheOtterlord (changes requested):
Found a few things that should be looked at

baklava/package.json
yummyweb: yes i am in case they want to build msi

yummyweb: there is a `npm run build:exe` script which builds exe.

baklava/package.json
yummyweb: I believe that should be decided later, since it worked in your fork. We could find a fix or something.

baklava/package.json
```diff
@@ -12,6 +12,7 @@
     "build:all": "npm run compile && electron-builder -mwl --win msi",
     "build:mac": "npm run compile && electron-builder --mac",
     "build:win": "npm run compile && electron-builder --win msi",
+    "build:exe": "electron-builder -p never --win",
```
theotterlord, line 15: why do you force no publish?

baklava/package.json
```diff
@@ -51,7 +52,10 @@
     },
     "win": {
       "icon": "icons/icon.png",
-      "target": "msi"
+      "target": [
+        "msi",
+        "exe"
```
theotterlord, line 57: I'm slightly worried about keeping the msi target. In my fork, the working action uses only exe so that msi is not built. If msi is built then we are adding a working build to a failing action meaning it won't get published. Removing msi for now is probably best. We can add it back later if we fix it

```suggestion
        "exe"
```

TheOtterlord:
Yeah, I just tested it locally. Leaving in `msi` as a build target will built `msi` and `exe`.

```
> dogehouse@1.0.14 build
> npm run compile && electron-builder


> dogehouse@1.0.14 compile
> tsc

  • electron-builder  version=22.10.5 os=10.0.19042
  • loaded configuration  file=package.json ("build" field)
  • writing effective config  file=builds\builder-effective-config.yaml
  • packaging       platform=win32 arch=x64 electron=11.3.0 appOutDir=builds\win-unpacked
  • Unpacking electron zip  zipPath=undefined
  • building        target=MSI arch=x64 file=builds\DogeHouse 1.0.14.msi
  • packaging       platform=win32 arch=ia32 electron=11.3.0 appOutDir=builds\win-ia32-unpacked
  • Unpacking electron zip  zipPath=undefined
  • building        target=nsis file=builds\DogeHouse Setup 1.0.14.exe archs=x64, ia32 oneClick=true perMachine=false
  • building block map  blockMapFile=builds\DogeHouse Setup 1.0.14.exe.blockmap
```

Issue #1223, opened 2021-03-19
Dogehouse and Profile Menu are push off screen on mobile devices when a room name is too long.


drvladb:
**Describe the bug**
When a room has a long name, it causes other buttons like the home button and profile picture to be moved off screen on smaller mobile devices (IPhone 6s or similar)

**To Reproduce**
1. Get phone with small screen
2. Open room with large name
or 
1. Create room
2. Shrink web browser to smallest size possible

**Expected behavior**
The text should be abbreviated so that the buttons are not offscreen 

**Screenshots**
![image](https://user-images.githubusercontent.com/33793511/111721491-5c1dcb00-8836-11eb-883b-b3e37f8382bd.png)

**What device are you on?**
(Screenshot) Chrome, Win10

daloes:
This was adressed in pr #1206 but apparently it's not working

daloes:
Confirmed issue persists
![Screenshot_20210319-083634_Chrome](https://user-images.githubusercontent.com/11330406/111746414-4fc55c80-888e-11eb-8a33-34130f66e55f.jpg)

MadanBhandari:
let me do this, and also user avatar is not vertically centered, will send pr for both the issues. someone assign this to me.

PR #1242, opened 2021-03-19
fix: long room name on navbar #1223

MadanBhandari:
Resolves: Dogehouse and Profile Menu are push off screen on mobile devices when a room name is too long. #1223


benawad:
does this mess up desktop?

MadanBhandari:
@benawad I have tested on desktop too, didn't notice any issues, no noticeable change on desktop.

for vertical alignment , align-items: center added
and for width, I have set it to take ~80%, which works fine for desktop and mobile

PR #1248, opened 2021-03-19
Grammatical error

neilgoyal:


sumnerevans:
I wonder how this happened in the first place... I assume it was copied directly from here: https://www.contributor-covenant.org/version/2/0/code_of_conduct/ or was created using the GitHub CoC template. It may be worth checking to ensure the entire thing is correctly copied.

PR #1257, opened 2021-03-19
custom emoji thugpepe

rishitxyz:
* Added a custom pepe emoji as thugpepe emoji of width 28px and height of 28px.


vi241:
you need to add the emoji to `emotedata.ts`

PR #1275, opened 2021-03-19
adds Esperanto translations

felipecespedes:
It will be good if someone else with Esperanto knowledge can check too, I did my best including Esperanto translations.

Dankon!

### DIFF
--- a/kofta/public/locales/eo/translation.json
+++ b/kofta/public/locales/eo/translation.json
@@ -1,0 +1,244 @@
+{
+  "_comment": "if you change this file, do: npm run i18",
+  "common": {
+    "loadMore": "ŝarĝi pli",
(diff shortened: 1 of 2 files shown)


ermalsh:
@benawad Esperanto does not have a language flag emoji, rather, based on online posts, it is recommended to use like a green square or a circle. 

This is completely different from our other languages in the selector, and will pop out more than the others, should we still allow for this type of flag emoji? I'm ok with it, just running it by you.

HarrisonMayotte (approved):
From a code standpoint it looks good, but if someone who knows Esperanto can confirm the translations make sense that'd be great.

overlisted:
might as well translate it in [interslavic](https://en.wikipedia.org/wiki/Interslavic_language) lmao

PR #1299, opened 2021-03-19
Fix undefined classname on buttons when classname is not passed as prop

redoper1:


wes337:
Could just be a default prop

HarrisonMayotte (approved):

redoper1:
> Could just be a default prop

Yeah, you are right. It's another possible solution that didn't come to my mind when I discovered this bug.

PR #1310, opened 2021-03-19
fix(roompage): user can view their avatar first as listener and speaker

### LINKED ISSUE #5: Have the own avatar always be on top / first in room
People are full of themselves, let them see themselves w/o searching

sam17896:
fix #5


HarrisonMayotte (approved):
Looks good to me.

PR #1311, opened 2021-03-19
feat(kofta): added 2 new doge emotes (28x28)

mitchglass97:
Added 2 doge emotes- Doge3D and DogeCool. both are gifs and both are 28px by 28px.


HarrisonMayotte (approved):
Looks doge. 👍🏼

PR #1322, opened 2021-03-19
Updated french translations 

EwanFox:
Added missing translations in french and update a few incorrect ones


HarrisonMayotte (approved):
Très bien! 👍🏼

PR #1327, opened 2021-03-20
Integrated storybook into Pilaf

Je12emy:
Integrated [Storybook](https://storybook.js.org/tutorials/intro-to-storybook/react-native/en/get-started/) into Pilaf, the app UI and the storybook UI can be switched with a `.env` file ([read more here](https://github.com/storybookjs/react-native#other-ways-to-render-storybook)) this is done with [react-native-config](https://github.com/luggit/react-native-config) but [setup for IOS](https://github.com/luggit/react-native-config#setup) is still needed

https://user-images.githubusercontent.com/34408108/111853467-c428eb80-88e0-11eb-8a97-cbba4d1881bf.mp4


benawad:
can you set it up so the stories can use TypeScript

Je12emy:
> can you set it up so the stories can use TypeScript

Will do!

Je12emy:
I've also integrated [storyloader](https://github.com/elderfo/react-native-storybook-loader#configuration) for automatically importing all stories, but it seems to display an error but only when configured to output a `.ts` file (loading stories still works), I can add this commit to this PR or for a different one, it's up to you.
image

pilaf/storybook/stories/index.ts
```diff
@@ -0,0 +1,2 @@
+import "./Button/Button.stories";
+// import './Welcome/Welcome.stories';
```
HarrisonMayotte, line 2: Is this one supposed to be commented out?

HarrisonMayotte (commented):
Everything looks good. Just need clarification on the comment below before I can "approve."

pilaf/storybook/stories/index.ts
Je12emy: Yeah, for some reason `welcome.stories` does seem to cause trouble so I commented it out just so it wouldn't be displayed in the navigation (this persisted on another fresh rn app).

HarrisonMayotte: Alright. Does it give any errors or anything?

Je12emy: The given component by storybook is class based, I guess TS doesn't like how it's written (The errors are not shown in JS).
image

Etienne-io: Welcome.stories is just a demo stories. You can remove all generated js stories and add one tsx stories as example.

image

HarrisonMayotte: Got ya.

HarrisonMayotte (approved):

Issue #1324, opened 2021-03-19
Atomic Components Implementation v3 (Mobile)


ajmnz:
<h1>Atomic Components v3 (Mobile)</h1>

### New components

To be implemented under `kibbeh/src/ui/mobile`.

- [x] MobileHeader - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2250
    - Requires modified UserAvatar > SingleUser ✅ 
- [x] FeaturedRoomCardAvatars - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2338
    - Requires none  ✅
- [x] FeaturedRoomCardHosts - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2327
    - Requires FeaturedRoomCardAvatars ❌
- [x] MobileNav - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2440
    - Requires none ✅
- [x] AccountOverlay - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2492
    - Requires none ✅
- [x] RoomCardParticipants - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2404
    - Requires modified UserAvatar > MultipleUsers ❌

###  Existing Components - Responsiveness

- [x] SearchBar - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2269
    - Requires SearchBar ✅
- [x] SearchResult > RoomSearchResult - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2354
    - Requires  SearchResult > RoomSearchResult ✅
- [x] SearchResult > UserSearchResult - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2388
    - Requires SearchResult > UserSearchResult ✅
- [x] SettingsIcon - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2449
    - Note the bottom border
    - Requires SettingsIcon ✅
[...]

PR #1328, opened 2021-03-20
Optimized the SearchBar for mobile

goldyydev:
A bit of a surprise for me that i've been able to work at 3am.

See #1324 for more info

### DIFF
--- a/kibbeh/src/ui/Search/SearchBar.tsx
+++ b/kibbeh/src/ui/Search/SearchBar.tsx
@@ -5,20 +5,22 @@ import { Input } from "../Input";
 export interface SearchBarProps
   extends React.ComponentPropsWithoutRef<"input"> {
   inputClassName?: string;
+  mobile?: boolean;
 }
 
 export const SearchBar: React.FC<SearchBarProps> = ({
   className = "",
   inputClassName = "",
+  mobile = false,
   ...props
 }) => {
   return (
     <div
-      className={`w-full bg-primary-700 text-primary-300 focus-within:text-primary-100 rounded-lg ${className}`}
+      className={`w-full bg-primary-700 text-primary-300 focus-within:text-primary-100 rounded-lg ${mobile ? 'px-4' : ''} ${className}`}
     >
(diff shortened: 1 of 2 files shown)


HarrisonMayotte (approved):

kibbeh/src/ui/Search/SearchBar.tsx
```diff
@@ -5,20 +5,26 @@ import { Input } from "../Input";
...
   inputClassName = "",
+  mobile = false,
   ...props
 }) => {
   return (
     <div
-      className={`w-full bg-primary-700 text-primary-300 focus-within:text-primary-100 rounded-lg ${className}`}
+      className={`w-full bg-primary-700 text-primary-300 focus-within:text-primary-100 rounded-lg ${
+        mobile ? "px-4" : ""
+      } ${className}`}
```
banky, line 21: Hey there! I think this component might be easier to use if instead of passing in a `mobile` prop, you use tailwinds built in stuff for checking for mobile screens. 

```suggestion
      className={`w-full bg-primary-700 text-primary-300 focus-within:text-primary-100 rounded-lg px-4 md:px-0 ${className}`}
```

This way it "automatically" knows it's on a phone vs desktop

kibbeh/src/ui/Search/SearchBar.tsx
```diff
@@ -5,20 +5,26 @@ import { Input } from "../Input";
...
     >
-      <div className="h-full mx-4 flex items-center pointer-events-none">
-        <SmSolidSearch />
-      </div>
+      {!mobile && (
+        <div className="h-full mx-4 flex items-center pointer-events-none">
+          <SmSolidSearch />
+        </div>
+      )}
```
banky, line 27: Similar to above, this can be 

```suggestion
        <div className="h-full mx-4 flex items-center pointer-events-none hidden md:block">
          <SmSolidSearch />
        </div>
```

goldyydev:
I wouldn't say that, it's a component; not the entire app. the handling component should set a mobile tag as we don't know if it will just be used for mobile.

PR #1331, opened 2021-03-20
feat: added 3 new emotes - Darth Kermit, Angry Kermit, and Anime Girl…

AlexWang18:
… Kermit


HarrisonMayotte (changes requested):
Please make sure you add the emojis to `kofta/src/app/modules/room-chat/EmoteData.ts` following the format you can see with all of the other emojis in there. 

Additionally, you should probably swap out the current image files for ones that are 28x28 to fit in with everything else and minimize file size for those on limited or metered connections. (This can be done with a simple image resizing tool - I'd recommend using http://www.simpleimageresizer.com/ and selecting "dimensions" and inputting the 28x28.)

AlexWang18:
I think I've fixed them, thanks for letting me know!

HarrisonMayotte (approved):
Looks good! Thanks for making the change!

PR #1358, opened 2021-03-20
Created MobileNav Component

bansalarnav:
Created MobileNav component (https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A2440)


goldyydev (approved):
Seems alright to me 👍

HarrisonMayotte (approved):
All good to me.

ajmnz (changes requested):
~~Mobile-only components should be put under `ui/mobile`, as stated in the issue.~~

All the icons are already available under `src/icons`. Some of them will need to be `20px` wide/tall (check on Figma), so you can pass `width={20} height={20}` as props to the icon component for it to scale to the desired size.

ajmnz (changes requested):
~~The existing `SmSolidPlus` needs to be left as is, because it's being used by `UpcomingRoomsCard`. You'll need to put the new plus icon under `icons/mobile` and `svgr` it.~~

HarrisonMayotte (approved):
Looks good to me.

HarrisonMayotte:
Not what I meant to do 😂. Just looking for a second review from @ajmnz

ajmnz (changes requested):
Pull latest from `staging`. Some of your icons aren't updated

goldyydev:
Can you please update the Storybook so it would work by default?

bansalarnav:
Closing this due to some problem with commits. Creating a new one (#1382 )

PR #1370, opened 2021-03-20
Adds AccountOverlay mobile component

sahilmob:


goldyydev (changes requested):
If you go on [this](https://www.figma.com/proto/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=4300%3A65&scaling=min-zoom) mode you can see that the AccountOverlay is supposed to be a **draggable** overlay (with the handle at the top). you have made the div itself but did not make it draggable nor is it currently an overlay.

I'd suggest you use something like react-draggable for this.

sahilmob:
Yes it was clear for me that it is draggable, however, I focused only on the UI part of the component.
Anyway I will make draggable and update the PR.

goldyydev (changes requested):
You took probably the most annoying component out there. hopefully this is the last fix 🤞 

image

sahilmob:
@goldyydev No problem at all :smile:

HarrisonMayotte (approved):

kibbeh/src/ui/mobile/AccountOverlay.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const AccountOverlay: React.FC<AccountOverlyProps> = ({
+  children,
+  position,
+  bounds,
+  dragHandler,
+  closeHandler,
+  className = "",
+  axis = "both",
```
goldyydev: Just a minor fix, this wouldn't be used for any xy dragging.
```suggestion
  axis = "y",
```

goldyydev (changes requested):
Just a minor fix

goldyydev (approved):

Issue #1369, opened 2021-03-20
Remove Outline When Buttons Clicked


ghost:
When buttons are clicked on the website the default blue outline appears , I find it quite annoying so could you remove , here is how you could do it:
``
button:focus{
outline:none;
}
``

AadityaMunjal:
I'm working on it

HonbraDev:
The outline is for users navigating with their keyboards. I would not remove it.

f1shy-dev:
Possibly we add focus classes that could change button colours/custom rounded outline instead? Making it look better instead of removing it is a better solution in my opinion.

>  here is how you could do it: `button:focus{ outline:none; }`

Also, DogeHouse uses TailwindCSS so we would do `button {@apply focus:outline-none}` to apply it to all buttons, or `<button className="focus:outline-none ..."` for a single button instead.

AadityaMunjal:
> Also, DogeHouse uses TailwindCSS so we would do `button {@apply focus:outline-none}` to apply it to all buttons, or `<button className="focus:outline-none ..."` for a single button instead.

I did that

benawad:
https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

vi241:
@benawad you can add better looking custom outlines or on focus background ...

PR #1374, opened 2021-03-20
Remove outline when button are clicked issue: #1369

AadityaMunjal:
#1369


benawad:
https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

PR #1380, opened 2021-03-20
doge house 🐕🏠...

Max-Hitchings:


🐕.🐕
```diff
@@ -1 +1 @@
-🐕
+🐕🐕
```
goldyydev: we're DogeHouse, not DogeDoge.
```suggestion
🐕🏠
```

goldyydev (changes requested):

mmattbtw:
You could also change the file name to `🐕.🏠`, if you wanted to.

vi241 (approved):

vi241:
Merge? 👍 / 👎

vi241 (approved):

vi241 (approved):

theCow61:
Originally the file was a gigabyte of doges but GitHub wouldn't let me push it because of a file limit. We either go big or small and because we can't go big with the gigabyte of 🐕 we should keep it small and simple with a singular 🐕 emoji in the file.

vi241:
tg it didn't

vi241:
that would be a big problem

PR #1382, opened 2021-03-20
Add MobileNav Component

bansalarnav:
Created MobileNav component


goldyydev:
Can you please update the Storybook so it would work by default?

bansalarnav:
It errors out as the component uses useRouter but it isn't available in storybook. If you have any workaround for it please tell so that I can make it work

bansalarnav:
Storybook works now. Please inspect the page and reduce page width to see the component as I added a media query to only show the component if width < 640px

goldyydev (approved):

goldyydev (changes requested):
A bit of an update, ben has removed the Sm prefix from icons, so please follow this PSA on this commit so we won't have weird names flowing around

image

bansalarnav:
> A bit of an update, ben has removed the Sm prefix from icons, so please follow this PSA on this commit so we won't have weird names flowing around
> 
> image

I already changed that. Icons don't have the Sm prefix

HarrisonMayotte (approved):

PR #1386, opened 2021-03-20
(Pilaf) Figma design

Etienne-io:
Improving values to get closer to Figma design
Removing unused style and import


pilaf/src/pages/LandingPage.tsx
```diff
@@ -7,47 +7,6 @@ import { colors } from "../constants/dogeStyle";
...
-        headers: {
-          "my-custom-header": "my custom header value",
-        },
-      });
-    } else Linking.openURL(url);
-  } catch (error) {
-    Alert.alert(error.message);
-  }
-};
-
```
HarrisonMayotte: Why is this all removed?

Etienne-io: The code was copied in the githubSignin component but wasn't remove from here

HarrisonMayotte: Alright

PR #1399, opened 2021-03-20
Kousa: Use simpler control flow structures where possible

melnary:
This replaces `cond` statements with a single branch and a `true` with a simpler `if` statement and `with` statements with a single statement with a simpler `case`.
This patch also removes a bunch of credo refactoring complaints relating to both of these.

Functionally there's no difference between the variants, but the code is more easily readable this way.


HarrisonMayotte (approved):
Looks good to me.

PR #1425, opened 2021-03-21
added translation for Croatian 

JakovGlavac:


nadirabbas:
Please add it in the LanguageSelector.tsx too

JakovGlavac:
It is already added 
https://github.com/benawad/dogehouse/blob/97e7eef9d29fb03d6e993f63d1a8a6f3321b745a/kofta/src/app/components/LanguageSelector.tsx#L22

nadirabbas:
oh my bad

PR #1440, opened 2021-03-21
Fixed translation mistakes

dk-raw:
**fixed some translation mistakes under `kofta/public/locales/el-GR/translation.json`**


jamesql:
looks good

jamesql (approved):

PR #1441, opened 2021-03-21
Migrate kofta to yarn

bansalarnav:
All the other modules are using `yarn` now. So no point of keeping `kofta` on `npm`


jamesql:
I think it was planned to move everything over to yarn, so good PR

benawad:
people are used to npm for kofta and this folder is going to die soon

PR #1446, opened 2021-03-21
Added BANNED emote

ShahrozTanveer:
added BANNED emote
BANNED


vi241:
We dont need more emotes rn because we have a lot but thank you for this pr

PR #1453, opened 2021-03-21
Better mute sound effects

HonbraDev:
*bonk*


vi241:
I will review this later

okdargy:
This is worse in my opinion.

vi241:
I think the same

PR #1471, opened 2021-03-22
Change browsers require interaction msg

bansalarnav:
I changed the `Browsers require interaction before they play audio` message to `Are you sure you want to join {room.name}`. This is more user friendly and also solves the purpose


kofta/src/webrtc/components/AudioRender.tsx
```diff
@@ -74,8 +85,7 @@ export const AudioRender: React.FC<AudioRenderProps> = () => {
       >
         <div className={`p-8 rounded m-auto bg-simple-gray-3c`}>
           <div className={`text-center mb-4`}>
-            Browsers require user interaction before they will play audio. Just
-            click okay to continue.
+            Are you sure you want to join {room?.name}?
```
amitojsingh366: I feel this might make some users think it's a hassle to click join two times. Whereas the previous one clearly stated it's their browsers fault and they need to interact with it once.

amitojsingh366: Also i think adding this to the translations would be a good idea too.

amitojsingh366 (approved):

kofta/src/webrtc/components/AudioRender.tsx
amitojsingh366: I approved by mistake so this comment is for requesting changes

amitojsingh366 (changes requested):

kofta/src/webrtc/components/AudioRender.tsx
bansalarnav: This is how most platforms like Google Meet, Zoom etc solve the problem of requiring interactions. Browser Interaction message is not as user friendly

amitojsingh366: @ArnavBansal11 that was just my opinion. If other people approve even ill approve but still, add the translation keys. I think this is a important thing that needs to be translated

bansalarnav: Ok i will add translation keys, but I will have to keep them in English for now due to my limited knowledge in linguistics

amitojsingh366: Yea sure!

bansalarnav:
added the translation keys

amitojsingh366:
> 
> 
> added the translation keys

Conflicts...

bansalarnav:
> 
> 
> > added the translation keys
> 
> Conflicts...

fixed :+1:

kofta/src/webrtc/components/AudioRender.tsx
jamesql: @amitojsingh366 this would only happen when a user joins through a link no?

jamesql (changes requested):
Should there be a ? at the end of the translation sentence

kofta/src/webrtc/components/AudioRender.tsx
bansalarnav: yep this will only happen then, not if a user joins through the list

bansalarnav:
> 
> 
> Should there be a ? at the end of the translation sentence

@jamesql i added the `?` in the react component itself as i couldn't understand how do interpolate 😅

jamesql (approved):

amitojsingh366 (approved):

kofta/src/webrtc/components/AudioRender.tsx
amitojsingh366: @jamesql yes that's correct, or might even happen if they're in a room and then open the site from another device *(throws you back into the room)*

bansalarnav:
@amitojsingh366 Can you please merge this? As there are conflicts every time a translation file is updated.... And i need to fix them

amitojsingh366:
> @amitojsingh366 Can you please merge this? As there are conflicts every time a translation file is updated.... And i need to fix them

Unfortunately not many people like this idea so ill be closing this

PR #1474, opened 2021-03-22
Fix error in kebab

bansalarnav:
Quick fix for a type error in `kebab` (error 
image
given below)

### DIFF
--- a/kebab/src/raw.ts
+++ b/kebab/src/raw.ts
@@ -17,20 +17,14 @@ export type Logger = (
   fetchId?: FetchID,
   raw?: string
 ) => void;
-export type ListenerHandler<Data = unknown> = (
-  data: Data,
-  fetchId?: FetchID
-) => void;
+export type ListenerHandler = (data: unknown, fetchId?: FetchID) => void;
 export type Listener = {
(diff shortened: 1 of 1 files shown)


kebab/src/raw.ts
```diff
@@ -107,7 +98,7 @@ export const connect = (
 
               listeners.push(listener);
 
-              return () => listeners.splice(listeners.indexOf(listener), 1);
+              return () => listeners.splice(listeners.indexOf(listener));
```
overlisted, line 101: This would unsubscribe all listeners after the needed one

kebab/src/raw.ts
```diff
@@ -17,20 +17,14 @@ export type Logger = (
   fetchId?: FetchID,
   raw?: string
 ) => void;
-export type ListenerHandler<Data = unknown> = (
```
overlisted, line 20: Are you sure the `Data` generic isn't used in wrapper?

overlisted (changes requested):
I have no idea how the error got there in the first place

overlisted:
And you didn't have to remove the generic...

PR #1488, opened 2021-03-22
Kousa: Remove starter processes

melnary:
This removes `Onion.StartRooms` and `Onion.StartRabbits` in favor of starting the necessary processes without intermediaries. The two GenServer modules did not keep any state and were only used to start other processes after the supervisor tree has been set up, becoming useless afterwards.

The code for starting the processes is now inside the parent Kousa module, but if anyone knows a better place for it, please tell me!


benawad:
this looks good to me, @ityonemo what do you think?

PR #1492, opened 2021-03-22
Fixes #1479 - Blur Chat button after opening

### LINKED ISSUE #1479: Holding the 'Open chat' button repeats the action
**The bug**
Pressing and holding the enter key while the "Chat" button is focused flashes the screen quickly between chat being open and closed. This is especially a problem for users who suffer from seizures. Note that pressing space while the button is focused on has the correct behaviour.

This bug is worse at narrow viewports, i.e. when the chat occupies the whole screen.

**Steps to Reproduce**
1. Enter a room
2. Tab to the "Chat" button
3. Hold down enter
  → :x: The chat flashes on and off rapidly

**Expected behaviour**
The chat will enable/disable once each time I hit enter

**Attatchments**
https://user-images.githubusercontent.com/50421330/111961358-e0977480-8ae8-11eb-8036-213ec713c62c.mp4

**What device are you on?**
Windows 10. Affects the Electron App (v1.0.16), staging.dogehouse.tv, and dogehouse.tv.

wes337:
After opening the chat, blur the element. This will prevent the `onClick` event from being repeatedly fired if holding down the Enter key.

To test this, hold down the Enter key while the Chat button is focused.

Fixes #1479


bansalarnav (changes requested):
Can you please do this for `Mute/Unmute` button in `BottomVoiceControl` too? It also seems to have the same issue where it repeatedly toggles on pressing Enter

amitojsingh366 (approved):
Seems fine to me

benawad:
I'm not sure if I want this. Keeping the focus on the element lets you toggle it. If you hold down enter, it is your own fault.

wes337:
> 
> 
> I'm not sure if I want this. Keeping the focus on the element lets you toggle it. If you hold down enter, it is your own fault.

Yeah that's a good point. Although I do think the Chat window should focus when you activate it, similar to when modals are auto-focused when they appear.

PR #1495, opened 2021-03-22
[Kofta] Add Australian translation

nihaals:


kofta/src/app/components/LanguageSelector.tsx
```diff
@@ -70,6 +70,7 @@ export const LanguageSelector: React.FC<LanguageSelectorProps> = ({
 
     /* Other languages */
     { value: "en-PIRATE", label: "🏴‍☠️ Pirate" },
+    { value: "au", label: "🇦🇺 uɐᴉꞁɐɹʇsnⱯ" }, // Australian
```
nihaals: Not sure if it makes more sense to have the flag at the end instead of the beginning

amitojsingh366:
😂😂

amitojsingh366:
I see what you're tying to do here but all languages need to follow the [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) codes. And unfortunately `au` is an unassigned code there. Not sure if unassigned means we can use it here...

amitojsingh366:
> 
> 
> I see what you're tying to do here but all languages need to follow the [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) codes. And unfortunately `au` is an unassigned code there. Not sure if unassigned means we can use it here...

But i think if `en-PIRATE` is a thing then ig this can also be valid

nihaals:
> I see what you're tying to do here but all languages need to follow the [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) codes. And unfortunately `au` is an unassigned code there. Not sure if unassigned means we can use it here...

I think that's what the `Other languages` section is for, pretty much all of the languages there are similar to this one.

I did make sure I wasn't actually intruding on an ISO 639-1 code, but as you said it's not listed, so it should be safe to class this as an "other" language.

ermalsh:
Apps using i18n commonly use this for en-*: **en-AU**, en-CA, en-GB, en-IN, en-US.

amitojsingh366:
@nihaals are you still working on this?

nihaals:
@amitojsingh366 Yes sorry, haven't been able to update it yet

amitojsingh366 (approved):
Looks good!

PR #1497, opened 2021-03-22
Fix global keybinds on linux

StratusFearMe21:
I found a library for global keybinds called gkm that seems to work just as well as iohook, but also works on linux. It also seems to have very similar syntax since I didn't have to change very much to patch it in.


baklava/package-lock.json
```diff
@@ -0,0 +1,5966 @@
+{
```
amitojsingh366, line 1: You are re-introducing `package-lock.json` which isn't a good idea as we have switched over to yarn.

amitojsingh366 (changes requested):
Unfortunately we cannot use this. The [`gkm`](https://github.com/tomzx/gkm) library you have used and the repo i traced it back to https://github.com/tomzx/gkm clearly shows that this module isn't being maintained anymore. Moreover it even states that the end user needs to have some version of a Java Virtual Machine installed on their local machine and even set in their path *(which is a hassle might i add)*.
Here is the exact quote:
*"GKM depends on [JNativeHook](https://github.com/kwhat/jnativehook), which runs on Java. Thus you will need to have a JVM available and more importantly, java availble on your PATH."*.
Hence we will not be considering this as an option.

baklava/package-lock.json
amitojsingh366: this is just a side note. the big comment I've written above is still the decision.

amitojsingh366:
closing this now.

PR #1502, opened 2021-03-22
Kebab: Mediasoup-client and small patches

overlisted:
Check out the new example (`mediasoup-audio`) for audio usage. Also, there is no Node support yet, but it will be added soon.


overlisted:
anyone wanna review this or do you just trust my code

PR #1503, opened 2021-03-22
Show all usernames in monospace (kofta)

overlisted:
Ben had approved this (privately)


overlisted:
And just saying I'm adding this because there's another overlisted but he replaced the l with an uppercase i, and we're now indistinguishable without monospace and I think that's not good

JackCrispy:
Ugly asf.

PR #1508, opened 2021-03-23
Actually ignore the .idea folder

samarmohan:
This did not work otherwise.


overlisted:
keep `*/.idea`

amitojsingh366:
> 
> 
> you should keep `*/.idea`

amitojsingh366:
> 
> 
> you should keep `*/.idea`

@overlisted let's close this then?

samarmohan:
It didn’t work otherwise. I tested it.

HarrisonMayotte:
Add both lines.

samarmohan:
I'll do that, the OG line ignored `.idea` folders in subfolders, not the root,

samarmohan:
You may need a `/` after `*/.idea`.

samarmohan:
> you should keep `*/.idea`

That doesn't work while my change works so there is no point in keeping that.
My change also ignores all `idea` folders in subdirs

HarrisonMayotte:
It does work- just not for the way you were using it. If someone opened a specific subfolder with the IDE that generates the .idea folder, it'll generate that in the subfolder, which is what the line there is for. ***Even if it didn't*** (which it does), one extra line for a "just in case" doesn't hurt.

samarmohan:
Ok, I’ll add it!

overlisted (approved):

PR #1509, opened 2021-03-23
Update README.md to reflect AUR presence

StratusFearMe21:
I have created a package for Dogehouse on the Arch User Repository which is an open space to post packages for the Linux distribution Arch Linux. It's best to assume that as long as the app works on Ubuntu, it will also work on Arch. So you shouldn't have to worry about Arch specific problems. It can be accessed [here](https://aur.archlinux.org/packages/dogehouse/). Currently I have a program running on my Raspberry Pi that will ensure that this entry in the AUR stays updated, similarly to how the github actions releases the builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,15 @@ A desktop app built with electron is available for Windows and Mac
 
 Download links are in [the releases section](https://github.com/benawad/dogehouse/releases/latest)
 
+Dogehouse is also available in the AUR
+```bash
+yay -S dogehouse
+```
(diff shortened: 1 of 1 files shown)


StratusFearMe21:
btw if you want me to take the package down or transfer ownership to someone on the dev team, I am totally open to doing those things

StratusFearMe21:
Updated the AUR package to 1.0.21 btw https://aur.archlinux.org/packages/dogehouse/

PR #1522, opened 2021-03-23
Baklava: Translations

TheOtterlord:
Someone on the Discord mentioned in the Baklava channel that they would like translations for the auto-updater notification. As this was not possible (we don't have access to the text), I replaced that system with event listeners that update the splash screen. giving us a discord-like updater. I've added localization using `i18next`, but for now only `en` has been added.


ermalsh:
This looks good to me, we may need to add a script for translations like we have in `kofta`, `npm run i18`

amitojsingh366 (approved):
This is a great feature!

TheOtterlord:
> This looks good to me, we may need to add a script for translations like we have in `kofta`, `npm run i18`

what does the script do?

amitojsingh366:
> > This looks good to me, we may need to add a script for translations like we have in `kofta`, `npm run i18`
> 
> what does the script do?

Syncs all the diff language translations

yummyweb:
@TheOtterlord please fix the conflicts first

yummyweb (commented):
It would be awesome if you would gamify it and add stuff like, "Hunting updates..." or "Cooking you updates.."

yummyweb (commented):
You removed the text it shows when not updating. Can you put the text back since there should be something there when the app is not updating.

TheOtterlord:
> You removed the text it shows when not updating. Can you put the text back since there should be something there when the app is not updating.

It shows `Launching...`

TheOtterlord:
conflics fixed. so do I need to write a script to sync language files, or do you think the ones in kofta are compatible?

TheOtterlord:
Fixed conflicts

amitojsingh366 (approved):

ermalsh:
> conflics fixed. so do I need to write a script to sync language files, or do you think the ones in kofta are compatible?

The ones in Kofta will work as long as they're generating the jsons for baklava too. So they may need to be updated.

TheOtterlord:
@ermalsh I don't know kofta very well, but I had a go anyway. I first tried copying and modifying the scripts, but it caused a lot of problems with typescript and was very messy. Now I'm going to try modifying the existing files in kofta to also generate for baklava as you said.

TheOtterlord:
Fixed conflicts. @ermalsh I can't seem to get a working version of the kofta stuff. If it's just generating a file with all keys as types, I'm not sure it's really that useful for baklava, but if we need to sync the translation files by essentially replacing missing keys with the english (default) values, I can code something up in the morning

amitojsingh366:
@TheOtterlord have a look at the 3 files here https://github.com/benawad/dogehouse/tree/staging/kofta/scripts
This is how kofta does it

TheOtterlord:
Thanks. I've added the files and generated the `translationKeys.ts` file

PR #1532, opened 2021-03-23
More stuff in kebab

overlisted:
- Better READMEs
- Cleaned up the examples
- Fixed a couple of audio related bugs and warnings
- Add the currently played user ID to consumer player parameters

Mic producing is still broken; don't merge if you need it

### DIFF
--- a/kebab/README.md
+++ b/kebab/README.md
@@ -2,4 +2,5 @@
 Client-side implementation of DogeHouse API.
 
 ## Usage
-Check out `examples/`
+**In web =>** see `examples/mediasoup-audio/` and `examples/react-chat`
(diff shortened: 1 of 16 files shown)


nadirabbas:
How will the `mediasoup-client` work for node? It is made for the browser. You could look into using `simple-peer` along with `wrtc` packages, which gives a backend implementation of WebRTC.

overlisted:
it won't work in node, im going to add a ffmpeg-based implementation later

nadirabbas:
`wrtc` is a WebRTC implementation for `node`

overlisted:
i know

PR #1535, opened 2021-03-24
[NEW FEATURE]: RPC for DogeHouse, DogePresence

yummyweb:
RPC for DogeHouse, dubbed, DogePresence is a feature like the one in Discord. When opened the desktop client, it automatically puts the currently opened app (except DogeHouse duh) on the profile for everyone to see. It is automatic and there is no way for the user to switch it off, but that will come soon.

PS (Only for Ben): Sorry for the new feature man.


TheOtterlord:
I think it would be better to add this once it can be toggled, otherwise users have no control over it except closing dogehouse when they use other apps if they don't want RPC. Also, as far as I can see, only the user can see there own RPC right now, so this seems very incomplete. Don't get me wrong, I'm not against the idea, but it seems this has a lot of implementation that is missing, and I think it would be best to complete the whole thing first.
I also think that it should wait until the new design (#1462)

yummyweb:
@TheOtterlord Toggling is almost done on the backend (electron process), but I would need someone to help me in the frontend. I know that new design should be completed and let it complete, till then I can add more features and add custom RPC like giving choice of which app to display like in Discord.

yummyweb:
> I think it would be better to add this once it can be toggled, otherwise users have no control over it except closing dogehouse when they use other apps if they don't want RPC. Also, as far as I can see, only the user can see there own RPC right now, so this seems very incomplete. Don't get me wrong, I'm not against the idea, but it seems this has a lot of implementation that is missing, and I think it would be best to complete the whole thing first.
> I also think that it should wait until the new design (#1462)

And everyone can see the RPC not only the user, @TheOtterlord

kofta/src/app/components/UserProfile.tsx
```diff
@@ -175,6 +176,7 @@ export const UserProfile: React.FC<UserProfileProps> = ({
             </div>
           );
         })}
+        Playing {useRPC()}
```
amitojsingh366, line 179: Won't this make it so the RPC only works for the user that has it on? Others wont be able to see it i assume

baklava/src/utils/keybinds.ts
```diff
@@ -142,7 +143,7 @@ export function RegisterKeybinds(bWindows: bWindowsType) {
             let PTT = PTT_STATUS.every((key_status) => key_status === true);
             if (PTT != PTT_PREV_STATUS) {
                 PTT_PREV_STATUS = PTT;
-                bWindows.main.webContents.send("@voice/ptt_status_change", !PTT);
+                mainWindow.webContents.send("@voice/ptt_status_change", !PTT);
```
amitojsingh366, line 146: Please revert this

amitojsingh366 (approved):

baklava/src/utils/keybinds.ts
amitojsingh366: I approved by mistake. I dont know how

amitojsingh366 (changes requested):

baklava/src/utils/keybinds.ts
yummyweb: Done

kofta/src/app/components/UserProfile.tsx
yummyweb: @TheOtterlord said the same thing. Isn't this component visible to everyone. Which is the visible UserProfile?

theotterlord: It is visible to everyone, but the data is coming from the user's computer, not the backend, so even if they look at someone elses account, it will show their RPC. You need some sort of storage on the backend to deal with this. This along with the new design I believe is a reason to wait for now, and plan out how this would work

amitojsingh366: Yes @TheOtterlord is right. I would say use the method that is being used to update the user's bio rn and make it a part of their bio

TheOtterlord:
I'm going to close this as the solution given does not match the feature. In addition, new features in Kofta are no longer allowed and new features are not currently accepted #1462 and a backend solution is also required for this feature.

kofta/src/app/components/UserProfile.tsx
yummyweb: Oh ok

PR #1536, opened 2021-03-24
Definitive global keybind fix

StratusFearMe21:
This fix uses a custom package written in Rust to capture global keyevents and send it off. I've updated the github actions, and the keybinds to correspond with this change. The only feature not implemented is customizable keybinds, which @amitojsingh366 said he would help me with


TheOtterlord (approved):
Looks good to me. I've also tested the gh action [here](https://github.com/TheOtterlord/dogehouse/runs/2182131696) and it worked fine.

PR #1539, opened 2021-03-24
Temporary bugfix for firefox

justmedev:
This is due to a instrument.ts bug. Till it gets fixed we have to stick to this solution or find a way around it. I am working on fixing it somehow

### DIFF
--- a/kibbeh/src/modules/webrtc/components/AudioRender.tsx
+++ b/kibbeh/src/modules/webrtc/components/AudioRender.tsx
@@ -63,7 +63,15 @@ export const AudioRender: React.FC<AudioRenderProps> = () => {
   const globalVolume = 100;
   const { consumerMap } = useConsumerStore();
   const audioRefs = useRef<[string, HTMLAudioElement][]>([]);
-
+  
+  // Check if browser is Firefox (if it is return)
+  // Because firefox currently doesnt work
+  if(typeof InstallTrigger !== 'undefined') {
+    // Return out of audioplayer (Audioplaying the way we do it doesnt work in firefox)
+    return <h1>Unsupported</h1><h5>This browser is currently not supported. We are working on it...</h5>
+  }
+  
+  // Else return the normal AudioRender
   return (
     <>
       <div


kibbeh/src/modules/webrtc/components/AudioRender.tsx
```diff
@@ -63,7 +63,15 @@ export const AudioRender: React.FC<AudioRenderProps> = () => {
   const globalVolume = 100;
   const { consumerMap } = useConsumerStore();
   const audioRefs = useRef<[string, HTMLAudioElement][]>([]);
-
+  
+  // Check if browser is Firefox (if it is return)
+  // Because firefox currently doesnt work
+  if(typeof InstallTrigger !== 'undefined') {
+    // Return out of audioplayer (Audioplaying the way we do it doesnt work in firefox)
```
amitojsingh366, line 70: It works fine for me...

amitojsingh366 (changes requested):
Can you be more clear? What exactly is the issue here? DogeHouse works fine for me on Firefox...

justmedev:
For me it just says 
`TypeError: n.play() is undefined` and the screen blacks out
And that happens in the `instrument.ts` lib.
So I just added this line to prevent the app from just making a black screen.

amitojsingh366:
But from the code you've added it looks like you made it so Firefox doesn't support audio at all...

justmedev:
> 
> 
> But from the code you've added it looks like you made it so Firefox doesn't support audio at all...

Well its just that firefox users know, that they should use chrome for dogehouse still the issue with instrument.ts is resolved

justmedev:
> 
> 
> But from the code you've added it looks like you made it so Firefox doesn't support audio at all...

Do you use your own build with the new design? I am using the current webversion maybe thats the problem

benawad:
Is this one a specific build of firefox?

amitojsingh366:
> 
> 
> > But from the code you've added it looks like you made it so Firefox doesn't support audio at all...
> 
> Do you use your own build with the new design? I am using the current webversion maybe thats the problem

No i don't use my own build. Here's my about page:
image

Share yours too so we can test it out

amitojsingh366:
> 
> 
> > But from the code you've added it looks like you made it so Firefox doesn't support audio at all...
> 
> Do you use your own build with the new design? I am using the current webversion maybe thats the problem

 No i don't use my own build. Here's my "about Firefox" page:
 image
 
 Share yours too so we can test it out

justmedev:
> 
> 
> Is this one a specific build of firefox?

I am using the latest standart edition: 87.0 (64-bit)

Heres a screenshot:
grafik

justmedev:
Oh I just tried it on Firefox Nightly (the same version @amitojsingh366 uses) and it works there. Seems to be a bug with Firefox 87.0
Thats the `instrument.ts` line thats causing it:
```js
// this fails for some browsers. :(
if (originalConsoleLevel) {
  Function.prototype.apply.call(originalConsoleLevel, global.console, args);
}
```
The comment was not added from me

melnary:
Can't reproduce this on Firefox 87.
Isn't `instrument.ts` a Sentry file anyway? Why would it be related to sound?

justmedev:
> 
> 
> Can't reproduce this on Firefox 87.
> Isn't `instrument.ts` a Sentry file anyway? Why would it be related to sound?

Hm.. Ok thats odd. Imma close this then

PR #1551, opened 2021-03-24
Updated hebrew translation

elad-weiss:


goldyydev:
@elad-weiss the entire Hebrew translation is in a male pronoun, maybe it will be worth it to keep it like that? (or change the whole thing)

elad-weiss:
I think it will be easier to only change this translation.
I'm closing this pull request and updating on a new one.

PR #1554, opened 2021-03-24
fix(kofta): realigns 'add to calendar' button back to original place …

### LINKED ISSUE #1460: Edit & Delete button for Scheduled Rooms pushes "Add to Calendar" button to the left
**Describe the bug**
When creating a scheduled room, you'll have the Edit & Delete button there because you own the scheduled room. But, the Add to Calendar button is moved to the side.

**To Reproduce**
1. Go to the Scheduled Rooms button
2. Schedule a test room or see an already scheduled room that you own
3. See error

**Expected behavior**
Make the "Add to Calendar" button below the Edit & Delete button

**Screenshots**
image

**What device are you on?**
Chromebook

Harry-Developer:
…on owned scheduled cards

When creating a scheduled event the 'add to calendar' button moves between date and edit, this
places it back to the right side of the card

fix #1460

Hi, first ever oepn source PR, please let me if anything needs changing and/or what I did wrong/could be improved. Thanks!

### DIFF
--- a/kofta/src/app/modules/scheduled-rooms/ScheduledRoomCard.tsx
+++ b/kofta/src/app/modules/scheduled-rooms/ScheduledRoomCard.tsx
@@ -81,17 +81,6 @@ export const ScheduledRoomCard: React.FC<ScheduledRoomCardProps> = ({
                 ? t("common.formattedIntlTime", { time: dt })
                 : t("common.formattedIntlDate", { date: dt })}
             </div>
-            <AddToCalendarButton
-              event={{
-                name: name,
-                details: description,
-                location: url,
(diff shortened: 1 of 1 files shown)


drvladb (approved):
Looks good!
image

samarmohan:
Looks good except the new design is out so this won't be needed.

drvladb:
@samarmohan I think that the old design will remain an option, and it is still supported because the new design is not yet ready (and is far from so). Plus, this is a fix, not a feature, so these are still reviewed.

ermalsh:
Looks good. We'll still keep this change even though new design is in prog.

PR #1565, opened 2021-03-25
Updated swiss language file

C0RR1T:
Just made the file better because the accent used was absolutely horrendous.

### DIFF
--- a/kofta/public/locales/gsw/translation.json
+++ b/kofta/public/locales/gsw/translation.json
@@ -75,7 +75,7 @@
       "requestingToSpeak": "Gsprächsafrage",
       "listeners": "Zuehörer",
       "allowAll": "Alles erlaube",
-      "allowAllConfirm": "Bisch der sicher? Das wird all {{count}} uffordere zum rede"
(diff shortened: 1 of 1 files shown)


kofta/public/locales/gsw/translation.json
```diff
@@ -63,43 +63,43 @@
...
+      "goHomeLinkText": "Hei gah"
     },
     "room": {
       "speakers": "Sprächer",
       "requestingToSpeak": "Gsprächsafrage",
       "listeners": "Zuehörer",
-      "allowAll": "Allow all",
-      "allowAllConfirm": "Are you sure? This will allow all {{count}} requesting users to speak"
+      "allowAll": "Alles erlaube",
+      "allowAllConfirm": "Bisch der sicher? Das wird all {{count}} uffordere zum rede"
```
CORR1T, line 78: watch out this is a wrong translation, correct one would be: "Bisch der sicher? Das wird all {{count}} Benutzer wo wennd rede Berechtigung gäh zum rede"

PR #1572, opened 2021-03-25
fix(roomdescription): add max height and scrolling to room description

### LINKED ISSUE #1556: Cannot read chat if the room description is too long.
**Describe the bug**
When the room description displays longer than the available height on the screen, it prevents a user from reading the chat, as they cannot scroll past the room description. 

**To Reproduce**
1. Create a room
2. Name it anything
3. For room description fill it with a sufficiantly long unicode character that will fill up the page. I used "𒐫".
4.  Open Chat by clicking chat on the bottom bar
5. Click "Show More" on the room discription
6. Now you are unable to view or read chat messages, and you are unable to scroll past the room description. 
**Expected behavior**
You are able to scroll past the room description 

**Screenshots**
image

**What device are you on?**
Chromebook

cursorial:
fix #1556

### DIFF
--- a/kofta/src/app/modules/room-chat/RoomDescription.tsx
+++ b/kofta/src/app/modules/room-chat/RoomDescription.tsx
@@ -23,7 +23,7 @@ export const RoomDescription: React.FC<RoomDescriptionProps> = () => {
 			<p className="text-gray-400 mb-1">
 				{t("modules.roomChat.roomDescription")}
 			</p>
-			<p className="whitespace-pre-wrap break-all">
+			<p className="whitespace-pre-wrap break-all overflow-y-auto max-h-24">
 				{expanded
 					? currentRoom.description
 					: truncatedString}


goldyydev:
I don't like 2 things about this PR
- You are adding a CSS file, this does not follow the convention we have
- You are using a static max-height

cursorial:
> I don't like 2 things about this PR
> 
> * You are adding a CSS file, this does not follow the convention we have
> * You are using a static max-height

Fair points, figured there would be some pushback on this, what is the convention for CSS? Are there any examples of commits I can look at?

Would % max-height be better?

cursorial:
> > I don't like 2 things about this PR
> > 
> > * You are adding a CSS file, this does not follow the convention we have
> > * You are using a static max-height
> 
> Fair points, figured there would be some pushback on this, what is the convention for CSS? Are there any examples of commits I can look at?
> 
> Would % max-height be better?

Updated PR to use tailwind classes instead of adding new CSS file.

Also now using `rem` instead of static `px` value for `max-height`.

goldyydev:
Alright, can you please show that it's working? (screenshot)

cursorial:
Screenshot from 2021-03-25 19-56-02
Screenshot from 2021-03-25 19-56-12
Screenshot from 2021-03-25 19-56-20

overlisted (approved):

PR #1581, opened 2021-03-25
fix(roomcard): hide long display names on roomcard

### LINKED ISSUE #1557: Long display names cause names to go past the edge of the page
**Describe the bug**
Very long display names continue outside the user list in the Room list.

**To Reproduce**
1. Change your display name to a name consisting of long Unicode characters. (i used ﷽)
2. Join a room
3. Click the house in the top left corner to go back to the page listing the rooms.

**Expected behavior**
The long display names should be truncated.
**Screenshots**
image

**What device are you on?**
Chromebook
**Additional context**

cursorial:
fix #1557

Before:
Screenshot from 2021-03-25 20-49-09

After:
Screenshot from 2021-03-25 20-46-39

### DIFF
--- a/kofta/src/app/components/RoomCard.tsx
+++ b/kofta/src/app/components/RoomCard.tsx
@@ -39,7 +39,7 @@ export const RoomCard: React.FC<RoomProps> = ({
     previewNodes.push(
       <div
         key={p.id}
-        className={`text-left text-simple-gray-d9 ${!i ? "mt-1.5" : "mt-0.5"}`}
+        className={`text-left text-simple-gray-d9 overflow-x-hidden ${!i ? "mt-1.5" : "mt-0.5"}`}
       >
         {p.displayName?.slice(0, 50)}
       </div>


benawad:
is it better to hide instead of word break thing?

cursorial:
Do we prefer one over the other?

benawad:
I think word break so it doesn't hide legit long descriptions

cursorial:
Cool, I'll update this tomorrow with that change.

Can I just clarify what you mean by

> word break thing

?

Do we want `...` at the end of the truncated value?
Do we want `see more`?
If we do want a `see more`, how will it expand? Down, across, scroll?

This has made me think, what if we adopt something similar to the way spotify displays text that is too wide for the screen where it marquees?

https://i.gyazo.com/76a39555b1cf12a35331537f06fbaa97.mp4

benawad:
oh you know what, I was thinking this was the description, but this is just someone with a really long name. I think I'm cool with just hiding it,

PR #1585, opened 2021-03-25
feat(global): Added .devcontainer for an easier local development environment

Drakota:
While trying to contribute on the project and seeing the many steps to build the project locally, I thought I'd make it easier for new contributors by making a [devcontainer](https://code.visualstudio.com/docs/remote/devcontainerjson-reference) that works with VSCode and setup everything automatically with Docker.

You need to have the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension installed:
image

When opening the repo you'll get the following notification:
image

By clicking Reopen in Container this will create a new devcontainer with Elixir, Node, Postgres and RabbitMQ installed.
In `devcontainer.json`
We can set all the ports we want to expose to our local machine (I may be missing some)
```json
"forwardPorts": [
        80,
        3000,
        4001,
        8080,
        5432,
        5672,
        15672
],
```

We can also specify VSCode extensions that will be automatically installed when the container is created, which is a good way to enforce to use extensions like ESLint and Prettier by everyone (I put the ones that were obvious, but we can add more).
```json
"extensions": [
        "dbaeumer.vscode-eslint",
        "esbenp.prettier-vscode",
        "jakebecker.elixir-ls",
        "bradlc.vscode-tailwindcss"
],
```
image

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,30 @@
+# Update to pick an Elixir version: 1.9, 1.10, 1.10.4
+ARG VARIANT=latest
+FROM elixir:${VARIANT}
+
+ENV USERNAME=doge
+ENV USER_UID=1000
+ENV USER_GID=$USER_UID
+
+COPY scripts/*.sh /tmp/scripts/
+
+# Create non-root doge user
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && bash /tmp/scripts/user.sh "$USERNAME" "$USER_UID" "$USER_GID" \
+    # Clean up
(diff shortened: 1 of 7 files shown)


benawad:
I like this idea, do you want to add this info to the contributing guide getting started section as a way to setup the backend

Drakota:
@benawad Yeah sure, I've added instructions for the devcontainer setup in CONTRIBUTING.md

cursorial:
This is really nice

amitojsingh366 (approved):
I can see this helping a lot!

Drakota:
Great, indeed I think this will help new contributors get setup fast and this also works well with [Github's codespaces](https://github.com/features/codespaces) feature. If we ever need to change the version of Elixir or Node we can also do that in the `Dockerfile`.

PR #1609, opened 2021-03-26
Add and fix some pt-pt translations

miguelcarreiro:
This PR includes:
- Translation of the non-translated pt-pt translation keys;
- Fixes some pt-pt translation keys;
- Execution of `npm run i18` command.


amitojsingh366:
omg why is this PR so big

miguelcarreiro:
> omg why is this PR so big

@amitojsingh366 That's because I ran `npm run i18` command.

PR #1634, opened 2021-03-27
Removed duplicate md size variant

BronzW:
There were duplicate online indicator for md size variant in SingleUser.tsx so I removed one.


benawad:
closed in favor of another

PR #1658, opened 2021-03-29
Kousa/isaac/eliminates room session gen registry

ityonemo:
step 1 in making kousa gen_servers more OTP compliant


ityonemo:
plz double check - i'm slightly unsure of that the join_room contract is satisfied; i feel like it *should* be returning userId but instead it returns a full user struct.

benawad:
the frontend needs to entire user data

benawad:
we could make it return just the user id but then every client will just make another request to get user details

PR #1683, opened 2021-03-30
deprecates GenRegistry for RoomChat genserver

ityonemo:


kousa/lib/onion/room_session.ex
```diff
@@ -37,8 +37,8 @@ defmodule Onion.RoomSession do
            Onion.RoomSessionDynamicSupervisor,
            {__MODULE__, Keyword.merge(initial_values, callers: callers)}
          ) do
-      {:error, {:already_started, pid}} -> {:ok, pid}
-      x -> x
+      {:error, {:already_started, pid}} -> {:ignored, pid}
```
ityonemo, line 40: @benawad decided to go with {:ignored, pid}, we might still want the pid somehow, this is more in line with the way that you use :ignore with GenServers (especially in the distributed genserver context).

kousa/lib/broth/socket_handler.ex
```diff
@@ -152,13 +152,12 @@ defmodule Broth.SocketHandler do
                       # @todo this should probably go inside room business logic
                       room = Rooms.get_room_by_id(user.currentRoomId)
 
-                      {:ok, room_session} =
-                        Onion.RoomSession.start_supervised(
-                          room_id: user.currentRoomId,
-                          voice_server_id: room.voiceServerId
-                        )
+                      Onion.RoomSession.start_supervised(
```
ityonemo, line 155: solution is to just not expect {:ok, session} LOL

PR #1692, opened 2021-03-31
fixes an issue where if you're not in a room, querying for the users …

ityonemo:
…fails

### DIFF
--- a/kousa/lib/beef/access/users.ex
+++ b/kousa/lib/beef/access/users.ex
@@ -64,6 +64,7 @@ defmodule Beef.Access.Users do
 
   def get_users_in_current_room(user_id) do
     case tuple_get_current_room_id(user_id) do
+      {:ok, nil} -> {nil, []}
       {:ok, current_room_id} ->
         {current_room_id,
          from(u in User,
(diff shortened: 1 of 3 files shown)


kousa/lib/beef/access/users.ex
```diff
@@ -64,6 +64,7 @@ defmodule Beef.Access.Users do
 
   def get_users_in_current_room(user_id) do
     case tuple_get_current_room_id(user_id) do
+      {:ok, nil} -> {nil, []}
```
ityonemo, line 67: here was the problem; if it came back with `{:ok, nil}`, it would pass nil into the succeeding sql query; querying with x == nil is no bueno.

kousa/test/broth/calls/get_current_room_users_test.exs
```diff
@@ -76,7 +76,6 @@ defmodule KousaTest.Broth.GetCurrentRoomUsersTest do
     @tag :skip
     test "returns what if the room doesn't exist"
 
-    @tag :skip
     test "returns what if you're not in a room", t do
```
ityonemo, line 79: as a bonus, this test is now activated.

PR #1700, opened 2021-03-31
room session hotfix

ityonemo:


ityonemo:
resolved using the ol' noggin

PR #1706, opened 2021-04-01
Update README.md to reflect an Ubuntu PPA

StratusFearMe21:
I figured that since @amitojsingh366 would like to disable updating on Linux (https://github.com/benawad/dogehouse/pull/1690), then an Ubuntu repository would be critical. I couldn't figure out for the life of me how to use launchpad.net unfortunately, however I did figure out how to self-host my own PPA very easily with my RPi at home. So I updated the README to include this new way to install dogehouse

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ Dogehouse is also available in the AUR
 ```bash
 yay -S dogehouse
 ```
+And as an Ubuntu ppa
+```bash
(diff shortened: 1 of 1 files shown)


StratusFearMe21:
Also, if someone could make a ppa.dogehouse.tv sub-domain and make an A record pointing to 100.15.73.250. We could make the PPA appear more legitamite than just https://ppa.palomagit.ml

PR #1709, opened 2021-04-01
Update PPA in README for http instead of https

StratusFearMe21:
I know I JUST got my previous pull request merged, but I've switched the server to use http instead of https, this is standard for Ubuntu repositories and I didn't realize that until recently. Security across the wire is handled by gpg  in ubuntu, not HTTPS. I'm sorry I didn't realize this the first time

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,8 @@ yay -S dogehouse
 And in an Ubuntu ppa
 
 ```bash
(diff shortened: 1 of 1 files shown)


amitojsingh366:
ok this checks out https://help.ubuntu.com/community/Repositories/Ubuntu

PR #1727, opened 2021-04-02
Added room subscription stuff

HonbraDev:
GitHub Desktop and me both had a stroke so I had to do it in the web version

### DIFF
--- a/kebab/src/wrapper.ts
+++ b/kebab/src/wrapper.ts
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-nocheck because internet is unpredictable
 
 import {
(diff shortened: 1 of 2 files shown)


kebab/src/wrapper.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-nocheck because internet is unpredictable
```
overlisted, line 2: are you sure typescript will detect the `// @ts-nocheck` now that it's on the second line

overlisted (approved):

kebab/src/wrapper.ts
HonbraDev: Yes, it worked. Added because ESLint kept screaming at me.

HonbraDev (commented):
Very nice :cement:

PR #1729, opened 2021-04-03
remove font size change - too small

dhenson02:


ajmnz:
We're not going to change the base `font-size` for now.

dhenson02:
its impossible to read need to at least set it to 16px standard
what concerns do you have about this?

ajmnz:
> its impossible to read need to at least set it to 16px standard
> what concerns do you have about this?

A value between 14-16px is the standard, combined with a readable font such as Inter (the one we are using). Check which font-size is using this platform, for example. You don't seem to have any problem to read my comments.
Screenshot 2021-04-03 at 13 01 07

dhenson02:
I see your "standard readable text" with this ...
image

ajmnz:
Which part isn't readable?

dhenson02:
The chat mostly and the names on the avatars

ajmnz:
> The chat mostly and the names on the avatars

Then start with that instead of changing the global `font-size`. Usernames in avatars and chat use `text-sm`, which is `0.875rem` or `~12px`, and that's why you see them small. Usernames might stay like that, but I designed the chat to be at 14px (even though there will be a chat setting to increase or decrease the `font-size`).

dhenson02:
you set it globally and it it doesnt just affect chat.  when i removed that prop the whole site had a subtle and very comfortable adjustment that was much more pleasing to read

lorstenoplo:
> you set it globally and it it doesnt just affect chat. when i removed that prop the whole site had a subtle and very comfortable adjustment that was much more pleasing to read

I kinda felt the same issue on the whole website the font aize is a bit small, little uncomfortable but if we increase it may spoil the whole look of the site. The font size change option should help to customise the size according to our eyes.

lorstenoplo:
image

Little big but a lot more comfortable to read

dhenson02:
This needs to reopen the font size on the site is too small. @benawad should weigh in here

Issue #1642, opened 2021-03-28
Started working on a Toki Pona translation.


GiggioG:
Toki Pona is a minimalistic conlang, and I wanted to translate DogeHouse into it. I am about 15% done at the point of writing this. I am writing this issue to inform anyone else that wants to translate DogeHouse into Toki Pona, that I have already started, but if you wanted to translate it, I would be glad to get some help. If you want to help, then reply to this in the bottom.

GiggioG:
I am using the `tp` locale, just saying if someone wants to use it, they can't.

PR #1786, opened 2021-04-04
feat(kofta): added translation to the language "Toki Pona" in the "tp…

GiggioG:
…" locale

feat #1642

Toki Pona is a philosophical artistic constructed language (or philosophical artlang) known for its small vocabulary. It was created by Canadian linguist and translator Sonja Lang for the purpose of simplifying thoughts and communication. - [Wikipedia](https://en.wikipedia.org/wiki/Toki_Pona)

### DIFF
--- a/kibbeh/public/locales/tp/translation.json
+++ b/kibbeh/public/locales/tp/translation.json
@@ -1,0 +1,267 @@
+{
+    "_comment": "if you change this file, do: npm run i18",
+    "common": {
+        "loadMore": "tenpo kama la sin li lon",
+        "loading": "sin li kama...",
+        "noUsersFound": "jan li ala",
+        "ok": "pona a",
+        "yes": "pona",
+        "no": "ike",
+        "cancel": "pini",
+        "save": "sitelen",
+        "edit": "ante",
+        "delete": "weka",
+        "joinRoom": "tawa tawa tomo",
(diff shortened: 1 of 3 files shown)


amitojsingh366:
according to [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) the `tp` language code is "Unassigned". so im not sure we can use this or not... I'll confirm this with others till then why not go and add the same translation to `kibbeh`?

amitojsingh366:
> 
> 
> according to [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) the `tp` language code is "Unassigned". so im not sure we can use this or not... I'll confirm this with others till then why not go and add the same translation to `kibbeh`?

@benawad is this allowed? assigning a custom language to an unassigned code i mean.

GiggioG:
it can always be changed, i just did tp because it was unassigned
and bruh, forgot about the kibbeh, ill start by seeing what can be copied over

GiggioG:
Did the kibbeh locale, didn't find the kibbeh version of `kofta/src/app/components/LanguageSelector.tsx` and the "Launguage" button on `next.dogehouse.tv` does nothing, so I didn't add the locale anywhere.

amitojsingh366:
ok seems good. lemme ask the others if using `tp` code is fine

PR #1802, opened 2021-04-05
Update globkey for new, faster approach (to fix heavy lag on Linux)

StratusFearMe21:
<details>
<summary>Long story (if you're intrested)</summary>
Before, globkey simply had 1 function called `getKeys()`, this function called a very expensive rust function every time it was called in js. It didn't create lag on Windows or Mac, but it did on Linux. So I had to find a way to make this work by only calling the expensive function once, and then using it for the rest of runtime. At first this was very difficult since the function couldn't be shared between threads, and in order to be global, it had to. Then, as I was about to open a question on Stack Overflow, @amitojsingh366 told me that he was going to use worker threads in order to implement the `getKeys()` function. Then I had a eureka moment, what if I just made a worker thread in rust, and then used mpsc for communication between rust and js.
</details>

TL;DR: globkey now works like this
- The `start()` function must be called when the app starts
- After that, calling the `getKeys()` function in a loop will get the global key events from the keyboard
- The `unload()` function must be called when the app exits **AND** the worker thread stops calling the `getKeys()` function, otherwise the app may freeze before closing (at least it did for me in testing).

(The functions that were there before still work, just when this PR gets merged they'll be properly deprecated)

### DIFF
--- a/baklava/src/utils/keybinds/worker.ts
+++ b/baklava/src/utils/keybinds/worker.ts
@@ -2,7 +2,6 @@ import { isMainThread, parentPort } from 'worker_threads';
 import globkey from 'globkey';
 
 if (!isMainThread) {
-    let prev_keys = ['', '']
     let shouldBreak = false;
     globkey.start();
     while (true) {
--- a/baklava/src/utils/keybinds/worker.ts
+++ b/baklava/src/utils/keybinds/worker.ts
@@ -4,12 +4,10 @@ import globkey from 'globkey';
 if (!isMainThread) {
     let prev_keys = ['', '']
     let shouldBreak = false;
+    globkey.start();
     while (true) {
         if (shouldBreak) break;
         let keys = globkey.getKeys();
-        if (keys != prev_keys) {
-            if (parentPort) parentPort.postMessage({ type: 'keys', keys: keys });
-        }
-        prev_keys = keys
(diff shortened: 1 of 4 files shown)


StratusFearMe21:
If you want me to change it back, feel free to tell me

amitojsingh366:
ill look into this in a few ~~hours~~ minutes from now.

baklava/src/utils/keybinds/worker.ts
```diff
@@ -2,14 +2,11 @@ import { isMainThread, parentPort } from 'worker_threads';
...
         let keys = globkey.getKeys();
-        if (keys != prev_keys) {
-            if (parentPort) parentPort.postMessage({ type: 'keys', keys: keys });
-        }
-        prev_keys = keys
+        if (parentPort) parentPort.postMessage({ type: 'keys', keys: keys });
     }
-}
\ No newline at end of file
+}
```
amitojsingh366, line 12: didn't you say you created a worker in rust? do we still have to use the worker i made in js and spent 7 hours to get working? *(please say yes, otherwise i will die)*

amitojsingh366 (changes requested):

baklava/src/utils/keybinds/worker.ts
amitojsingh366: nvm i tested this branch out locally and it works.

PR #1817, opened 2021-04-05
removed black outline around components

husseinamine:
why? 
cuz it looks better

edit: and also for user experience


overlisted:
@ajmnz what do you think

benawad:
this is needed for accessibility

ajmnz:
Yeah, but outline should only be visible on tab, not on click.

PR #1820, opened 2021-04-05
Try to fix MacOS sound

StratusFearMe21:
This PR updates globkey and attempts to fix the terrible sound produced by MacOS when using PTT, see https://imgur.com/a/o9VdGPo

### DIFF
--- a/baklava/src/utils/keybinds/worker.ts
+++ b/baklava/src/utils/keybinds/worker.ts
@@ -2,9 +2,13 @@ import { isMainThread, parentPort } from 'worker_threads';
 import globkey from 'globkey';
 
 if (!isMainThread) {
+    let prev_keys = ['', '']
     globkey.start();
     while (true) {
         let keys = globkey.getKeys();
-        if (parentPort) parentPort.postMessage({ type: 'keys', keys: keys });
+        if (keys != prev_keys) {
+          if (parentPort) parentPort.postMessage({ type: 'keys', keys: keys });
+        }
+        prev_keys = keys
     }
 }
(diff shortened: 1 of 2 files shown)


StratusFearMe21:
I also noticed that the `shouldbreak` logic that was here before was removed by someone. I know I didn't in my previous PR, but this have something to do with this issue

amitojsingh366:
> 
> 
> I also noticed that the `shouldbreak` logic that was here before was removed by someone. I know I didn't in my previous PR, but this have something to do with this issue

actually i was the one who removed the `shouldBreak` logic as i was the one who put it there but forgot to make it actually do the task of breaking the loop. instead i noticed `worker.terminate()` automatically breaks the loop for me so i didn't touch `shouldBreak` and forgot to remove it. basically `shouldBreak` was redundant.

baklava/src/utils/keybinds/worker.ts
```diff
@@ -2,9 +2,13 @@ import { isMainThread, parentPort } from 'worker_threads';
 import globkey from 'globkey';
 
 if (!isMainThread) {
+    let prev_keys = ['', '']
```
amitojsingh366, line 5: wait this line wasn't here? how did typescript not throw a bazillion errors!?

amitojsingh366 (approved):

StratusFearMe21:
okay

baklava/src/utils/keybinds/worker.ts
StratusFearMe21: before, rust handled the prev_keys logic and all that stuff, so this update hands that over to TS.

amitojsingh366: ahh makes sense...

PR #1821, opened 2021-04-05
Add FeaturedRoomCard component in Pilaf

kleo-xhindoli:
* Adds `FeaturedRoomCard` component in Pilaf. It exposes the same props as `RoomCard`.
* Updates `RoomCard` props to take in a list of `<Tag>` elements instead of `<Text>` so that we can support glowing tags

Screenshot:
image


BronzW:
lmao why does it look like you just cut off doge's nose

lorstenoplo:
Excluding the design I wish there could have been a prime category indicators with the tags.
What I meant to say is something like this : 
Screenshot_20210405-184432

@ajmnz can u weigh in here?

kleo-xhindoli:
> lmao why does it look like you just cut off doge's nose

The nose is propagated to the main doge leading the charge

Etienne-io:
@lorstenoplo Displayed information are based on figma design for the moment 
image

Update will be done at figma's pace

BronzW:
I don't think this is accurate with the figma version. I don't know if it's just me or is there more space between the 'Hosted by' and the 'Marcus Bloch, Don Velez' text? On the figma version they seem to be a bit closer together but on the screenshot of the pilaf version they seem more spaced.

ajmnz:
And it has a drop shadow too

BronzW:
This should be reopened and remerged once this is equivalent with the figma version.

Issue #1822, opened 2021-04-05
Use the uuid npm library for all uuid validation or generation vs. custom code


mic-max:
Replace "pilaf/src/lib/isUuid.ts" with npm uuid.validate() method

https://www.npmjs.com/package/uuid

iamanalt:
Is there any good reason to do this?

PR #1824, opened 2021-04-05
Use npm.uuid library vs. custom code

mic-max:
Addresses this "bug" #1822


overlisted (approved):

benawad:
If you delete it from `kibbeh`, you need to replace all of the instances

mic-max:
> If you delete it from `kibbeh`, you need to replace all of the instances

Oops, I forgot to hit save in my editor for that file, it should be fully replaced now

PR #1842, opened 2021-04-06
emotes work with lowercase & json is lowercase

theCow61:
Inconstant when typing out emotes and emojis so now with this it all gets converted to lowercase so you don't have to worry about accidentally using a upper case or not. Also re formats EmoteData so they are all lowercase names. @nadirabbas @rban123 helped on this.


HarrisonMayotte (approved):
Looks good to me!

theCow61:
How do we force people to use lowercase names when PRing emotes.

benawad:
@theCow61 we are not taking anymore emote prs

theCow61:
@benawad can we squeeze in this DoughDoge?
image

benawad:
@theCow61 yeah but kill 1 emote

PR #1847, opened 2021-04-06
Just a little baklava update

yigdo:
Making keywords for Baklava package.json

### DIFF
--- a/baklava/package.json
+++ b/baklava/package.json
@@ -23,7 +23,9 @@
     },
     "keywords": [
         "DogeHouse",
-        "Electron"
+        "Electron",
+        "voice-chat",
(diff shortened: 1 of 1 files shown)


amitojsingh366:
what exactly is the point of this?

baklava/package.json
```diff
@@ -23,7 +23,9 @@
     },
     "keywords": [
         "DogeHouse",
-        "Electron"
+        "Electron",
+        "voice-chat",
+        "Doge"
```
amitojsingh366, line 28: what exactly is the point of this?

amitojsingh366 (changes requested):

PR #1859, opened 2021-04-06
style:removes button outline when focused, instead replacing it with a change to the button's secondary color

fishsticks89:
style:removes button outline when focused, instead replacing it with a change to the button's secondary color

adds comments/documentation to global.css and button.tsx


ajmnz (approved):
100% against outlines on click.

benawad:
https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

fishsticks89:
This is a really good point

I'll make a new pull request that only removes the outline when it is focused and not tabbed to, using the focus-visible attribute

I made it change colors because I foresaw this problem, but in hindsight, for the visually impaired, this would not be sufficient.

PR #1860, opened 2021-04-06
DoughDoge emote

theCow61:
DoughDoge replacing animegirlpepe (Sorry whoever added that but I had to replace something randomly.)
Had to convert the doughdoge jpeg to a png with transparent background but without the background its hard to tell that its Dough Doge and could also be interpreted as Twinkie Doge so its multi purpose.

This is DoughDoge with transparent background that is being used in this commit. 
doughdoge


fishsticks89:
Why did you have to replace anime girl Pepe?

theCow61:
Was told to kill an emote if I was to put one in. ( #1842 )

fishsticks89:
Survival of the fittest 🤣

Issue #1863, opened 2021-04-07
Emotes with menu don't work after lower casing json.


theCow61:
**Describe the bug**
#1842 brakes emotes. Emote menu auto populates input field thing with old emote name which doesn't match with the update json.

Can be fixed by automatically lower casing the emote names when the menu auto populates the input field or reverting the #1842 push.

nadirabbas:
you need to convert to lowercase in `createChatMessage` too.

amitojsingh366:
i've fixed this. will push soon

theCow61:
Wierd the v.toLowerCase() isn't working right now but when I did it on my fix PR with yarn staging it worked.

theCow61:
Never mind its working now.

amitojsingh366:
@theCow61 it will only work after ben deploys it, and it seems like ben has deployed it now

PR #1869, opened 2021-04-07
Fixes emotes and also obama.

theCow61:
#1863 lowercases v in tokens in createChatMessage.ts which will make all emotes go out as lowercased even if you typed out with uppercases so it won't matter what the emote menu gets wrong on casing.

Also fixes obama emote, for some reason the emote menu was putting :obama: in the text field when :obamium: is needed to actually display emote, so I just changed name so it would match the short_name and now it will work with :obama:

### DIFF
--- a/kibbeh/src/lib/createChatMessage.ts
+++ b/kibbeh/src/lib/createChatMessage.ts
@@ -39,7 +39,7 @@ export const createChatMessage = (
     } else if (item.startsWith(":") && item.endsWith(":") && item.length > 2) {
       tokens.push({
         t: "emote",
-        v: item.slice(1, item.length - 1),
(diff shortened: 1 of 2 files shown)


amitojsingh366:
as i mentioned [here](https://github.com/benawad/dogehouse/issues/1863#issuecomment-814762999) I've already fixed this im just fixing another thing with baklava and will push my fixes after its done. the obamium -> obama change seems unnecessary

theCow61:
When you click it in the emote menu :obama: will show up and that isn't the correct word for it to display the emote.

amitojsingh366:
> 
> 
> When you click it in the emote menu :obama: will show up and that isn't the correct word for it to display the emote.

i didn't quite get you but go ahead and make a separate pr ill have a look into it soon

PR #1882, opened 2021-04-07
fix(kibbeh): redirect user to dash page if access token is present in the local storage

### LINKED ISSUE #1874: New version bug: doesn't re-authenticate user when you close and re-open app
**Describe the bug**
The current user doesn't get authenticated when you close the tab or visit another url and then open dogehouse again. It takes you back to the login page even when there's a token in localstorage.

**To Reproduce**
1. Go to 'next.dogehouse.tv'
2. Login
3. Close the tab or visit another url
4. Open dogehouse again

**Expected behavior**
User should be re-authenticated as long as there's a valid token in localstorage. 

**Screenshots**

**What device are you on?**

**Additional context**

sahilmob:
fix #1874


benawad:
this conflicts with some existing code, I will be fixing it soon

PR #1883, opened 2021-04-08
fix(kibbeh): fix input classnames spacing

### LINKED ISSUE #1867: The username's input field background changes to white when invalid
When creating a username the length of characters must be between 4 - 14 and they must only be alphanumeric/ underscore right? 
When the username is invalid, the background of the input field changes to white, and the color of the text changes to light grey. That's just too bright and I can't see the text properly. The new users might not be comfortable with it. 

It's better to keep the dark background and let the color of the text be either red or any bright color of your choice. 

**Additional context**
image

sahilmob:
It seems that the space at the end of the `cn` string is ignored if it is concatenated with other string.
fix #1867


benawad:
can you rebase your changes with the goldy commits removed

benawad:
nvm I'm just going to refactor this bit, its ugly

PR #1892, opened 2021-04-08
(feat. Kibbeh) The Language Selector.

goldyydev:
This PR is following @ajmnz's guidelines and awaiting @benawad's review (since this is a feature)


benawad:
@goldyydev remove framer motion dep and then you can merge, looks good

goldyydev:
Also removed emoji-mart as we're not using it anywhere in kibbeh.

PR #1895, opened 2021-04-08
reset chat store on route change start 

### LINKED ISSUE #1891: New Version Bug: Chat messages from first room are transferred when going into a new room.
https://user-images.githubusercontent.com/50546858/114031007-3a36c780-987b-11eb-93e1-27118b9e4f5a.mp4

sahilmob:
fix #1891


benawad:
If you go to the home page, but don't go into another room, I don't want the chat to be cleared. Chat should only be cleared upon joining a room you weren't already in.

PR #1903, opened 2021-04-09
Created Nynorsk translation

FranticBomb:


TheOtterlord:
Hello @FranticBomb!

Thanks for the pull request.
While new languages are accepted, they need to be in the same format as the others.
So they need to be in the file `/kofta/locales/<language-code>/translation.json` and use the same keys as the other languages.
Take a look at a previous translation pull request for reference.

PR #1907, opened 2021-04-09
Added Nynorsk translation

FranticBomb:


dk-raw:
We do not use kofta anymore, transfer the files to [kibbeh](https://github.com/benawad/dogehouse/tree/staging/kibbeh).

PR #1908, opened 2021-04-09
Added Nynorsk

FranticBomb:


dk-raw:
We do not use `kofta` anymore, transfer the file into [kibbeh](https://github.com/benawad/dogehouse/tree/staging/kibbeh)

PR #1913, opened 2021-04-09
replace multiple else/ifs with switch case

faisalsayed10:
Just a small contribution where I tried to replace those else-ifs with switch case which looks better in my opinion.

### DIFF
--- a/kibbeh/src/modules/layouts/DesktopLayout.tsx
+++ b/kibbeh/src/modules/layouts/DesktopLayout.tsx
@@ -23,35 +23,35 @@ export const DesktopLayout: React.FC<MainLayoutProps> = ({
 
   let middle = null;
   
-  switch(screenType) {
-		case "3-cols":
-			middle = (
+  switch (screenType) {
+    case "3-cols":
+      middle = (
       <>
         <LeftPanel>{leftPanel}</LeftPanel>
         {children}
         <RightPanel>{rightPanel}</RightPanel>
       </>
     	);
-			break;
-		case "2-cols":
-			middle = (
+      break;
+    case "2-cols":
+      middle = (
       <>
         <LeftPanel>{tabletSidebar}</LeftPanel>
         {children}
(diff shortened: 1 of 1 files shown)


kibbeh/src/modules/layouts/DesktopLayout.tsx
```diff
@@ -22,37 +22,44 @@ export const DesktopLayout: React.FC<MainLayoutProps> = ({
...
+			middle = (
       <>
         {children}
         {floatingRoomInfo}
       </>
-    );
+    	);
+			break;
+		default:
+			break;
```
overlisted, line 62: you don't need this
```suggestion
```

overlisted (changes requested):

kibbeh/src/modules/layouts/DesktopLayout.tsx
faisalsayed10: the default thing?

dhenson02 (changes requested):
project indents are spaces not tabs. please don't mix

whatever IDE you use, download the editorconfig plugin and enable it to make this easier for you

kibbeh/src/modules/layouts/DesktopLayout.tsx
```diff
@@ -22,37 +22,41 @@ export const DesktopLayout: React.FC<MainLayoutProps> = ({
   const screenType = useScreenType();
 
   let middle = null;
-
-  if (screenType === "3-cols") {
-    middle = (
+  
+  switch(screenType) {
```
dhenson02: consistency - statements in project are padding with space around parenthesis - `switch (` not `switch(`

dhenson02 (commented):
nothing major just something to keep in mind

kibbeh/src/modules/layouts/DesktopLayout.tsx
faisalsayed10: alright so I replaced the indentation with spaces

Issue #1929, opened 2021-04-09
Keybinds section should display key icons instead of text


marceliwac:
**Is your feature request related to a problem? Please describe.**
The current keybind descriptions are not looking great; they are basic and could be improved.

There are also few minor bugs with the displayed combination:
- The incorrect order of displayed key combination, i.e. `[Shift] + [9]` results in displayed **9 + Shift** rather than **shift + 9**.
- The displayed keys *sometimes* apply modifiers to the keys that are used (i.e. `[Shift] + [a]` will be displayed as **Shift + A** and not **Shift + a** (lowercase "a"). This isn't a huge problem for standard ASCII characters, however... setting the modifier such as `[alt]` followed by `[o]` results in sequence that produces "ó" on the combined character on keyboards with layouts that support them, rather than it's primitive (i.e. `[alt] + [o]` results in **alt + ó** rather than **alt + o**). This is not a huge deal, since the key combination works fine, but it impairs the legibility of the shortcut.

**Describe the solution you'd like**
 This could be easily improved with some additional CSS and a component surrounding the displayed keybinds.

**Describe alternatives you've considered**
Custom CSS for keybind look.

**Additional context**

---

<img width="475" alt="Screenshot 2021-04-09 at 20 20 19" src="https://user-images.githubusercontent.com/17514757/114231113-e955ca80-9971-11eb-90df-b43525b589b4.png">

---

<img width="1164" alt="Screenshot 2021-04-09 at 20 20 06" src="https://user-images.githubusercontent.com/17514757/114231124-ebb82480-9971-11eb-8624-b5f8ec61def4.png">

PR #1935, opened 2021-04-09
WIP: Visual keybind representation.

marceliwac:
The patch aims to add a visual representation of the set keybinds (as outlined in #1929)

At the moment the implementation is just a rough draft of what the outcome should look like. Some help would be needed to put the CSS in correct place and possibly target some issues with different modifier keys (ctrl, alt etc.)

# Help needed

### DIFF
--- a/kibbeh/src/modules/keyboard-shortcuts/ChatKeybind.tsx
+++ b/kibbeh/src/modules/keyboard-shortcuts/ChatKeybind.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 import { recordKeyCombination } from "react-hotkeys";
 import { useKeyMapStore } from "../../global-stores/useKeyMapStore";
 import { Button } from "../../ui/Button";
+import {VisualKeybind} from "./VisualKeybind";
 
 interface ChatKeybindProps {
   className?: string;
@@ -39,7 +40,7 @@ export const ChatKeybind: React.FC<ChatKeybindProps> = ({ className }) => {
       <div className={`ml-4`}>
         toggle chat keybind:{" "}
         <span className={`font-bold text-lg`}>
-          {active ? "listening" : CHAT}
+          {active ? "listening" : <VisualKeybind keybind={CHAT.toString()} />}
         </span>
       </div>
     </div>
--- a/kibbeh/src/modules/keyboard-shortcuts/InviteKeybind.tsx
+++ b/kibbeh/src/modules/keyboard-shortcuts/InviteKeybind.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 import { recordKeyCombination } from "react-hotkeys";
 import { useKeyMapStore } from "../../global-stores/useKeyMapStore";
 import { Button } from "../../ui/Button";
+import {VisualKeybind} from "./VisualKeybind";
 
 interface InviteKeybindProps {
   className?: string;
@@ -39,7 +40,7 @@ export const InviteKeybind: React.FC<InviteKeybindProps> = ({ className }) => {
       <div className={`ml-4`}>
         invite keybind:{" "}
         <span className={`font-bold text-lg`}>
-          {active ? "listening" : INVITE}
+          {active ? "listening" : <VisualKeybind keybind={INVITE.toString()} />}
         </span>
       </div>
     </div>
--- a/kibbeh/src/modules/keyboard-shortcuts/MuteKeybind.tsx
+++ b/kibbeh/src/modules/keyboard-shortcuts/MuteKeybind.tsx
@@ -3,6 +3,7 @@ import { recordKeyCombination } from "react-hotkeys";
 import { useKeyMapStore } from "../../global-stores/useKeyMapStore";
 import { useTypeSafeTranslation } from "../../shared-hooks/useTypeSafeTranslation";
 import { Button } from "../../ui/Button";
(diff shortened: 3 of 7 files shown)


BronzW:
What do you mean by putting css in the correct place?

BronzW:
Inline styling works but using tailwind is better as it is what we use for kibbeh

amitojsingh366:
ok ive said this before as-well but if you are thinking of making design change please confirm with @ajmnz beforehand, as if he/ben won't like it its gonna be rejected and all your time wasted.

kibbeh/src/modules/keyboard-shortcuts/ChatKeybind.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 import { recordKeyCombination } from "react-hotkeys";
 import { useKeyMapStore } from "../../global-stores/useKeyMapStore";
 import { Button } from "../../ui/Button";
+import {VisualKeybind} from "./VisualKeybind";
```
dhenson02, line 5: try to keep formatting consistent with the rest of the code - `import { VisualKeybind } ...`

kibbeh/src/modules/keyboard-shortcuts/InviteKeybind.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 import { recordKeyCombination } from "react-hotkeys";
 import { useKeyMapStore } from "../../global-stores/useKeyMapStore";
 import { Button } from "../../ui/Button";
+import {VisualKeybind} from "./VisualKeybind";
```
dhenson02, line 5: here too

dhenson02 (changes requested):
- keep consistent formatting with the rest of the code
- use i18n for any user-facing static text in the UI

kibbeh/src/modules/keyboard-shortcuts/ChatKeybind.tsx
```diff
@@ -39,7 +40,7 @@ export const ChatKeybind: React.FC<ChatKeybindProps> = ({ className }) => {
       <div className={`ml-4`}>
         toggle chat keybind:{" "}
         <span className={`font-bold text-lg`}>
-          {active ? "listening" : CHAT}
+          {active ? "listening" : <VisualKeybind keybind={CHAT.toString()} />}
```
dhenson02, line 43: why are all of these calling `.toString()`?  doesn't appear they will ever be returned as anything but a string already

dhenson02 (approved):

marceliwac:
Thanks for the feedback @BronzW @amitojsingh366 @dhenson02!

This PR at the moment serves more of a feature idea that presents a possible outcome on what the changes could look like, rather than the how they should be implemented (hence the WIP tag). I'm happy to acknowledge that @ajmnz is responsbile for the design, so I'll happily take their lead on what the UI should look like. I thought I'd address some of the feedback you guys have so here goes:

- **CSS implementation**
I'm not very familiar with tailwind CSS so the current code aimed just to show what the keybind visuals would look like. I'll happily implement this correctly via tailwind once @ajmnz signs off on the redesign; for now I just used the inline styles to mock it up.

- **`toString()` conundrum**
Again, perhaps through fault of my own, lack of experience in TypeScript led me to this dirty fix to get rid of some of the compilation errors that arose from the `KeySequence` type, which declared few different types. I didn't take time to solve it, seeing as it's more of a mockup at this stage.
This also brings me to a second point, which is described in the [linked issue](https://github.com/benawad/dogehouse/issues/1929): _the displayed keys sometimes apply modifiers to the keys that are used (i.e. [Shift] + [a] will be displayed as Shift + A and not Shift + a (lowercase "a")_. 
Depending on how we collectively decide to tackle this aspect, it will either need changing to some other type, which can then be used for `keybind` prop of the `VisualKeybind` component, or we can just get rid of `toString()` in whichever way will satisfy the compiler.

- **i18n**
I'm not sure how the displayed keys will differ between different languages, but I don't suppose this PR will warrant any changes on that front. It appears that the current version of the code is not using any additional formatting for the keybinds and in general, key codes (i.e. Control -> Ctrl) tend to be the same across languages.

- **Key acronyms**
Somewhat linked to the above, one of the parts I'm unsure about is what the keys should be abbreviated to. For example, the **meta** key on the Mac is a **cmd** (or command) key and **alt** is an **option**, but this might differ on other systems like Windows or Linux. It would be nice to adapt to those different systems depending on the keyboard layout/os combination that is being used, but I would have to look into whether the [react-hotkeys](https://www.npmjs.com/package/react-hotkeys) provides support for that. If so, that shouldn't be too difficult, otherwise we can either scratch it completely and just use the keybinds returned by the **react-hotkeys** or write some logic that inferes these (which will lead to a bit of a rabbit hole in terms of deciding which systems should this be tested on, how to get the user OS etc.).

Thank you all again for the feedback. I'm looking forward to hearing from @ajmnz and moving forward with this.

kibbeh/src/modules/keyboard-shortcuts/ChatKeybind.tsx
marceliwac: Please see my [comment](https://github.com/benawad/dogehouse/pull/1935#issuecomment-817143927) below.

ajmnz:
This is no longer needed (#2282)

Issue #1936, opened 2021-04-09
New Version Bug: Edit profile on other peoples profile !?!?!


ghost:
I see a button that says edit profile on other peoples profiles clicking it edits mine but still what?

PR #1937, opened 2021-04-09
Added so that only the specified user is able to see the edit profile, voice settings and sound settings buttons. 👌

BronzW:


BronzW:
Fix #1936

PR #1947, opened 2021-04-10
baklava/package.json: Add snap interfaces + fix spelling bug in NSMicrophoneUsageDescription

ajhalili2006:
### Quick fix an spelling bug in `NSMicrophoneUsageDescription`

```diff
// Tabs and spaces are removed for readibility
- "NSMicrophoneUsageDescription": "Mic access for speeach input in voice."
+ "NSMicrophoneUsageDescription": "Mic access for speech input in voice."
```

### Added configuration for Snap builds

While I can hear people from rooms with the `audio-playback`, I can't speak because snapd doesn't auto-connect `audio-record` by default (and also not defined in its `baklava/snap/snapcraft.yml`, in this case in `baklava/package.json`) which [requires to be requested to be auto-connected in the Snapcraft Forums](https://forum.snapcraft.io/t/process-for-aliases-auto-connections-and-tracks/455). So I added `audio-record` interface to the list.

https://user-images.githubusercontent.com/34998342/114271211-dced7280-9a42-11eb-97fa-30d6a97abc3b.mp4

## Additional Resources

* For documentation regarding configuring snap builds, see https://www.electron.build/configuration/snap.
* For interfaces, see https://snapcraft.io/docs/supported-interfaces and https://snapcraft.io/docs/interface-management


baklava/package.json
```diff
@@ -68,6 +68,21 @@
...
                 "snap"
+            ],
+            "category": "AudioVideo"
+        },
+        "snap": {
+            "confinement": "strict",
+            "plugs": [
+                "default",
+                "audio-playback",
+                "audio-record",
```
amitojsingh366, line 79: have you properly tested this locally? does it fix the mic problems?

amitojsingh366 (changes requested):

baklava/package.json
ajhalili2006: I'll try to do the build first, the install my build with the `--dangerous` flag (`snap install baklava/build<dogehouse-build-i-reproduced>.snap --dangerous` and check if it's work.

Because the `audio-record` interface isn't automatically connected due to privacy reasons as per the Snapcraft docs, I may manually connect with `snap connect dogehouse:audio-record` and test if it's detected.

amitojsingh366: yes please test it out.

ajhalili2006: @amitojsingh366 It works! (ref: https://open.lbry.com/screencast-devlog-1:b70f45420143972e1f6ea014c4a80cc17194911f)

Issue #1807, opened 2021-04-05
New Version Bug: Inaccurate/Untranslated Texts


Zamion101:
**Describe the bug**
While using new version of DogeHouse (https://next.dogehouse.tv), some text of the website are not translated and  rendered as `English` rather than `User Language`. This behaviour can be observed in `Profile Page` `https://next.dogehouse.tv/u/`. Also this issues can be related to #1799 .

**To Reproduce**
1. Go to https://next.dogehouse.tv/dash
2. Go to Profile page
3. Observe the issues

**Expected behavior**
Texts should be rendered in `User Language` rather than default behaviour (`English`). 

**Screenshots**
In the Screenshot below there are some translated text in the middle of the screen and un-translated texts on the sides. Also search bar contains un-translated texts.
![image](https://user-images.githubusercontent.com/8071263/113531350-d5544680-95c8-11eb-9955-0feb636b378f.png)

In the Screenshot below there is a console error which describes 404 HTTP Error for `locales/tr-TR/translation.json`
![image](https://user-images.githubusercontent.com/8071263/113531954-798abd00-95ca-11eb-89d7-29cb613a7337.png)

**What device are you on?**
Opera (74.0.3911.218) on Windows 10 Pro x64.

nadirabbas:
I don't think we have i18n in the next version yet.

gruselhaus:
I think a tool (CLI-Script, what so ever) that checks for missing translations would be great!

TheOtterlord:
Kibbeh now has translations. Most strings are translated, but if you are able to contribute translations, please make a pull request.

BronzW:
@TheOtterlord yes, but there are some hardcoded strings that haven't been converted into translation keys. Almost all new components doesn't have them, like the profile page. Instead of having this issue open we should make it a requirement to add translation keys whenever @ajmnz adds new components for people to create.

PR #1964, opened 2021-04-11
Added a bunch of missing translation keys to several components in Kibbeh

BronzW:
1. Added translation keys to 
- Following Online component
- Feed component
- Search Bar component
- Profile Card component
- Settings Dropdown component
- Upcoming Rooms Card component
2. Added `lowercase` className to force following and followers text to be lowercase as they should so translations don't mess that up
3. Removed `capitalize` on some places as they cause grammatical errors in other languages.


BronzW:
Fix #1807

PR #1973, opened 2021-04-12
Fixing FollowersOnline typo.

Giovanny-DS:
This component have been imported as FollowersOnline. I assume that is because the name is "FollowersOnline.tsx" but was exported as "FollowerOnline".
this was generating warnings with storybook, and others linters

### DIFF
--- a/kibbeh/src/ui/FollowersOnline.tsx
+++ b/kibbeh/src/ui/FollowersOnline.tsx
@@ -21,7 +21,7 @@ export interface FriendsOnlineProps {
   showMoreAction?: MouseEventHandler<HTMLDivElement>;
 }
 
-export const FollowerOnline: React.FC<UserWithFollowInfo> = ({
+export const FollowersOnline: React.FC<UserWithFollowInfo> = ({
   username,
   avatarUrl: avatar,
   online,


kibbeh/src/ui/FollowersOnline.tsx
```diff
@@ -21,7 +21,7 @@ export interface FriendsOnlineProps {
   showMoreAction?: MouseEventHandler<HTMLDivElement>;
 }
 
-export const FollowerOnline: React.FC<UserWithFollowInfo> = ({
+export const FollowersOnline: React.FC<UserWithFollowInfo> = ({
```
amitojsingh366, line 24: This component should just stay as `FollowerOnline` as it isn't a list of followers, rather it is a component containing a **single** follower's data *(username, avatar, etc)* and links to their profile and room they're in

amitojsingh366: also no to forget. you're not changing any imports in any other files. this will end up break the code.

amitojsingh366 (changes requested):

goldyydev:
This PR is not needed. The FollowerOnline name is correct.

kibbeh/src/ui/FollowersOnline.tsx
Giovanny-DS: > also no to forget. you're not changing any imports in any other files. this will end up break the code.

yeah, you're right. if you look at [kibbeh/src/stories/FriendsOnline.story.tsx](https://github.com/benawad/dogehouse/blob/a53dba92ecda6ec7f182a781edd3127e57a53537/kibbeh/src/stories/FriendsOnline.story.tsx). at line 7
it is begin imported as FollowersOnline. it is causing errors. that's the line that should be fixed

Giovanny-DS:
> This PR is not needed. The FollowerOnline name is correct.

yeah, you're right the name is correct. 
the incorrect import is on [kibbeh/src/stories/FriendsOnline.story.tsx](https://github.com/benawad/dogehouse/blob/a53dba92ecda6ec7f182a781edd3127e57a53537/kibbeh/src/stories/FriendsOnline.story.tsx). at line 7.

kibbeh/src/ui/FollowersOnline.tsx
Giovanny-DS: also, is correct to name this component as FollowersOnline when the export is FollowerOnline?.

PR #1979, opened 2021-04-12
style(createscheduledroommodal): the Date input functions as a button…

ryndm:
… but didn't have cursor-pointer

I've changed the cursor styling to pointer for Datepicker in CreateScheduledRoomModal. The input
functions as a button so I thought that it should have a cursor pointer


benawad:
use tailwind

PR #1993, opened 2021-04-13
Improved locale/cs

prokopschield:
I translated all the new strings that were recently added, + a few improvements

### DIFF
--- a/kofta/public/locales/cs/translation.json
+++ b/kofta/public/locales/cs/translation.json
@@ -4,9 +4,9 @@
     "loadMore": "Načíst více",
     "loading": "Načítám...",
     "noUsersFound": "Nikdo tu není",
-    "ok": "ok",
-    "yes": "ano",
-    "no": "ne",
+    "ok": "Ok",
+    "yes": "Ano",
+    "no": "Ne",
     "cancel": "Zrušit",
     "save": "Uložit",
     "edit": "Pozměnit",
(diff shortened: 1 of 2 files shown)


kofta/public/locales/cs/translation.json
```diff
@@ -138,23 +139,23 @@
...
     "keyboardShortcuts": {
-      "setKeybind": "nastavit klávesovou zkratku",
-      "listening": "posluchám",
-      "toggleMuteKeybind": "ztlumit mikrofon",
-      "togglePushToTalkKeybind": "Pust-To-Talk",
-      "toggleOverlayKeybind": "toggle overlay keybind"
+      "setKeybind": "Nastavit klávesovou zkratku",
+      "listening": "Poslouchám",
+      "toggleMuteKeybind": "Ztlumit mikrofon",
+      "togglePushToTalkKeybind": "Push-To-Talk (mluvit jen po zmáčknutí tlačítka)",
```
goldyydev, line 154:
```suggestion
      "mluvit jen po zmáčknutí tlačítka",
``` 

Shouldn't this be changed to this?

kofta/public/locales/cs/translation.json
```diff
@@ -138,23 +139,23 @@
...
+      "toggleOverlayKeybind": "Nastavit klávesovou zkratku pro overlay"
     },
     "userVolumeSlider": {
-      "noAudioMessage": "z nějakého důvodu se nepřehrává zvuk"
+      "noAudioMessage": "Z nějakého důvodu není vysílán zvuk..."
```
goldyydev:
```suggestion
      "noAudioMessage": "Z nějakého důvodu není vysílán zvuk"
```
what is this for?

goldyydev (changes requested):

kofta/public/locales/cs/translation.json
iamanalt: It shouldn't, one is not sufficiently descriptive without the other.

kofta/public/locales/cs/translation.json
iamanalt: no audio consumer for some reason

This happens when no audio consumer is spawned by MediaSoup on the backend, usually because the particular person isn't streaming any.

This usually means the person is a bot.

PR #1994, opened 2021-04-13
Updated the French

dk-raw:
**Updated the French translation with the new phrases**

### DIFF
--- a/kofta/public/locales/fr/translation.json
+++ b/kofta/public/locales/fr/translation.json
@@ -39,7 +39,7 @@
     "home": {
       "createRoom": "Créer une salle",
       "refresh": "Actualiser",
-      "editRoom": "Edit Room"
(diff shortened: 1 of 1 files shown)


kofta/public/locales/fr/translation.json
```diff
@@ -39,7 +39,7 @@
     "home": {
       "createRoom": "Créer une salle",
       "refresh": "Actualiser",
-      "editRoom": "Edit Room"
+      "editRoom": "Modifier la salle "
```
goldyydev, line 42: Why is there an extra space?
```suggestion
      "editRoom": "Modifier la salle"
```

goldyydev (changes requested):

BronzW:
Please also change the translations in /kibbeh

PR #2003, opened 2021-04-13
Updated Apple link to https

thiwup:
Pretty simple! Just changed the link from http:// to https://


amitojsingh366:
pretty sure this is gonna break it.

Issue #2000, opened 2021-04-13
Chat input doesn't clear when changing rooms


wes337:
**Describe the bug**
The chat input doesn't clear when leaving and entering a new room.

**To Reproduce**
1. Enter a room.
2. Type a message (but don't send it)
3. Leave room, and join another
4. Your unsent message stays

**Expected behavior**
Input should be cleared/reset when entering a new room.

**Screenshots**
![](http://g.recordit.co/92ayL0UpYY.gif)

**What device are you on?**
Windows 10, Firefox

PR #2008, opened 2021-04-13
fix(roomchatinput): clear chat input when leaving room

Mayowa-Ojo:
prevent chat input from leaking to other rooms by clearing the input when user leaves current room

**fix** #2000

This is a fix for the referenced issue above. Currently when you leave a room while typing a message, the message remains persisted when you join another room.

By adding a hook to clear the input when `RoomChatInput` component unmounts, the issue is fixed.


benawad:
I'm fixing this in a different way, because I want this to keep the text if its the same room

Mayowa-Ojo:
Alright.

Issue #1996, opened 2021-04-13
Baklava doesn't have a custom title bar


BronzW:
The current one clashes with the DogeHouse theme by being white.

PR #2010, opened 2021-04-14
baklava: added custom titlebar

amitojsingh366:
- added custom titlebar on bakalva
- below is a recording of how it looks:

https://user-images.githubusercontent.com/35400192/114663537-2ec42000-9d03-11eb-91ae-3d91c1d6c5f5.mp4

### DIFF
--- a/kibbeh/src/modules/layouts/ElectronHeader.tsx
+++ b/kibbeh/src/modules/layouts/ElectronHeader.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+import { WinMaximizeIcon, WinCloseIcon, WinMinimizeIcon } from "../../icons";
+export function ElectronHeader() {
+  return (
+    <div className="electron-header">
+      <div className="header-image-cont">
+        <img
+          className="header-image"
+          src="https://github.com/benawad/dogehouse/raw/staging/kibbeh/public/img/doge.png"
+          width="20px"
+          height="20px"
+        />
+      </div>
+      <p className="header-title">DogeHouse</p>
+      <div className="header-icons">
+        <div className="header-icon">
+          <WinMinimizeIcon />
+        </div>
+        <div className="header-icon">
+          <WinMaximizeIcon />
+        </div>
+        <div className="header-icon">
+          <WinCloseIcon />
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/kibbeh/src/modules/layouts/ElectronHeader.tsx
+++ b/kibbeh/src/modules/layouts/ElectronHeader.tsx
@@ -1,5 +1,5 @@
 import isElectron from "is-electron";
-import React from "react";
+import React, { useState } from "react";
 import {
   WinMaximizeIcon,
   WinCloseIcon,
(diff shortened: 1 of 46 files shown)


BronzW:
Maybe the x should be 'accent' color when it is hovered?

TheOtterlord:
Closes #1996

TheOtterlord (approved):

amitojsingh366:
> Maybe the x should be 'accent' color when it is hovered?

done

kibbeh/src/modules/layouts/ElectronHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Button } from "../../ui/Button";
+
+let ipcRenderer: any = undefined;
+if (isElectron()) {
+  ipcRenderer = window.require("electron").ipcRenderer;
+}
+
+export function ElectronHeader() {
+  return (
+    <div className="electron-header z-50">
```
BronzW, line 26: `user-select: none` is added to custom headers as a norm as it is not part of the normal UI but part of the application

BronzW: Also adding `-webkit-app-region: drag;` would allow us to drag the window wherever we want.

BronzW (changes requested):
Changes necessary to make titlebar act as a real titlebar.

BronzW:
Fixing the scrollbar popping up when the header is present would also be good as it currently pops up to the side because of the header, which is not how it is supposed to act.
image

kibbeh/src/modules/layouts/ElectronHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+        />
+        <Button
+          icon={<WinMinimizeIcon width={10} height={10} />}
+          size="small"
+          color="secondary"
+          className="remove-outline"
+          onClick={() => {
+            if (isElectron()) ipcRenderer.send("@app/minimize");
+          }}
+        />
```
BronzW: Using the UI button is not meant for things like this. Using the normal `<button>` will get rid of the rounded edges (titlebars usually don't have them) and it will make it more dynamic and easier to change.

amitojsingh366: > Fixing the scrollbar popping up when the header is present would also be good as it currently pops up to the side because of the header, which is not how it is supposed to act.
> image

why should the scrollbar not be present?

amitojsingh366:
here's how it looks right now. both mac and windows shown:

https://user-images.githubusercontent.com/35400192/114739730-cfd9c780-9d51-11eb-9304-a70a484c8cf3.mp4

kibbeh/src/modules/layouts/ElectronHeader.tsx
BronzW: Because without the titlebar the scrollbar isn't there, so why should it be there when it's present?

amitojsingh366:
latest design:

https://cdn.discordapp.com/attachments/812993958112133170/831996160456917093/2021-04-14_23-54-43.mp4

amitojsingh366:
latest design:

https://user-images.githubusercontent.com/35400192/114778157-2c9ea780-9d7d-11eb-9727-dd7cedafa15d.mp4

PR #2017, opened 2021-04-14
Updated Lithuanian translation

siuglukas:
Question:  What's the point of running "npm run i18" when l update the translation?


BronzW:
Running `yarn i18` on other languages than english is not needed.

PR #2028, opened 2021-04-15
Updated a bunch of translations

dk-raw:
**Added translation for eh new phrases on 7 different languages**


BronzW:
Are you a native speaker of these languages or do you have experience with them before? Becuase just running them through google translate is not the best thing.

amitojsingh366:
do you know all of these languages or are you just google translating them all?

dk-raw:
Yes, I got help from a couple of friends.

PR #2029, opened 2021-04-15
Updated LanguageSelector and corrected language codes

rzsxdev:
- Corrected some languages codes (sr-Latin -> sr-LATIN)
- Fixed labels and comments in LanguageSelector
- Added or modified dome emojis that were black flags
- Corrected the order so that everything is alphabetical except, for example, the German languages (de, de-AT, gsw) which are grouped together

**Tagalog and Basque are missing translation files (Also Nynorsk)**
Additionally, do you think it would be better if the languages are ordered like in Wikipedia with languages with Latin script first, then Cyrillic script, and so on?


dk-raw:
`grc` is not a thing. ancient Greek is referred to as `ag-GR`

rzsxdev:
I checked from [Wikipedia](https://en.wikipedia.org/wiki/Ancient_Greek) and other sources that Ancient Greek's language code was `grc`, but I did not find anything called `ag-GR`. However, I may be mistaken.

dk-raw:
Being Greek myself, I have never heard of that language code again. `ag-GR` is the official code for the Ancient Greek

dk-raw:
Correction. `grc` is referred to as Greek until 1453(whick was a pre-mature version of the modern Greek). The pure Ancient Greek (1000 B.C.) is `ag-GR`.

rzsxdev:
Ok, I'll correct it then.

dk-raw:
~~use `agr` for the Ancient Greek. It is more suitable~~
Edit:
use `grc` if you are following the ISO pattern. However I have never seen that before.

rzsxdev:
So i'll keep it as it is?

dk-raw:
> So i'll keep it as it is?

Keep it as it is If you are following the ISO standards for all languages.

PR #2034, opened 2021-04-15
dev-env: Add Gitpod config

ajhalili2006:
This PR contains configuration and required tools to be installed during prebuilds for Gitpod SaaS and self-hosted. It also includes an `docker-compose.yml` (adapted from `.devcontainer/docker-compose.yml`, with only RabbitMq and Postgres included) to ensure you don't need to mess up with local development (AKA experience dependency hell), among other things.


amitojsingh366:
what exactly is gitpod tho?

ajhalili2006:
> what exactly is gitpod tho?

@amitojsingh366 Gitpod is an dev-environment-as-code service, looks like GitHub Codespaces, [but better.](https://www.gitpod.io/gitpod-vs-github-codespaces)

overlisted:
why does this ide need a special config to work with a project

ajhalili2006:
> why does this ide need a special config to work with a project

@overlisted While Gitpod can work out of the box without the configuration (`.gitpod.yml`), the problem is this project is some sort of monorepo (Erlang and Elixir are not pre-installed in `gitpod/workspace-full` image), so there's an custom `.gitpod.Dockerfile` (and of course, `docker-compose.yml` that I copied from `.devcontainer/docker-compose.yml` with only Postgres and RabbitMQ containers needed).

ghost:
what's the point of gitpod when GitHub Codespaces exists?

overlisted:
> what's the point of gitpod when GitHub Codespaces exists?

it doesn't yet (afaik)

ajhalili2006:
Since I marked this as `ready for review` (removing the `WIP: ` prefix for GitLab users), you can test if the proposed Gitpod config in this PR works.

[Try it here](https://gitpod.io#github.com/MadeByThePinsHub/dogehouse), and let me know if anything breaks.

Speaking of `docker-compose.yml` file, please do `docker-compose up -d` manually as this is not yet tested.

benawad (changes requested):
merge conflcits

overlisted:
i could've solved them if i knew a little bit more docker

PR #2044, opened 2021-04-15
Changed the selector color

dk-raw:
**Made the selected text become white with a black background. Made the change under `kofta/src/index.css`**

### DIFF
--- a/kofta/src/index.css
+++ b/kofta/src/index.css
@@ -63,3 +63,8 @@ option {
   background: none;
   border: none;
 }
+
+::selection {
+  background-color: black;
+  color: white;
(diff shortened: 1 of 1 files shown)


kofta/src/index.css
```diff
@@ -63,3 +63,8 @@ option {
   background: none;
   border: none;
 }
+
+::selection {
+  background-color: black;
```
amitojsingh366, line 68: i believe @benawad doesn't want custom selector rn. but he said that a while ago so i may be wrong

amitojsingh366 (changes requested):

PR #2053, opened 2021-04-16
updated hebrew translation

elad-weiss:


TheOtterlord (approved):

TheOtterlord:
oops. I did not realise this changed Kofta. 
@elad-weiss if you would like to update translations, please do so in Kibbeh

PR #2083, opened 2021-04-17
refactor(kibbeh): match search overlay to with design spec

Mayowa-Ojo:
This PR:

- fixes the current issue with the search overlay
- doesn't affect the responsive behaviour of the search components AFAIK

To get the intended result:
- the SearchBar is now placed as a sibling of the search overlay instead of a child component
- the search overlay is hidden by default and only becomes visible when open/focused

This is what it looks like now:

Screenshot from 2021-04-16 18-26-35


ajmnz (approved):

ajmnz (changes requested):
https://user-images.githubusercontent.com/57961822/115281089-ec5a7300-a148-11eb-9bf6-bea12769b423.mov

Results shouldn't overlap the search bar.

Also, spacing on the x-axis between the search bar and the overlay should be 10px (right now is kind of off with 9px on the left and 14px on the right). Also, that spacing should apply to search results too.

Mayowa-Ojo:
I'm currently working using latest changes from the **staging** branch and the search functionality doesn't work. I can't see search results because it errors out for some reason.

what branch has a working search bar?

Mayowa-Ojo:
Ok, nvm. Worked around that. I think it's fixed now.

ajmnz (changes requested):
Screenshot 2021-04-20 at 21 09 20

Nope, it's overlapping even more and the SearchResult should have the same width as the search bar.

Mayowa-Ojo:
Fixed.

Issue #2022, opened 2021-04-14
[Profile Page | Desktop] Components Implementation


ajmnz:
# Components Implementation

### Profile Page | Desktop

New components available for anyone that wants to contribute to the new design (`kibbeh`). Make sure to read the guidelines at the bottom of the issue.

---

### New components
- [x] ProfileHeaderWrapper - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A977
    - Requires none ✅ 
- [ ] ProfileHeader - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A978
    - Requires ProfileHeaderWrapper,  UserBadge (for *Follows you*), Button, UserAvatar/SingleUser ❌
- [x] ProfileTabs - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A989
    - Requires none ✅
- [x] UserBadgeLg - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A990
    - Requires none ✅
- [x] ProfileAbout - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A1048
    - Requires UserBadgeLg ❌
- [x] BaseDropdownSm
    - Includes:
       - CalendarDropdown - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A1049
       -  ShareDropdown - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A1054
    - Requires none ✅
- [x] UpcomingRoomCardLg - https://www.figma.com/file/CS01VVLR7ArQl0afYFkNj3/Web-App?node-id=8277%3A1062
    - Requires CalendarDropdown, ShareDropdown, UserAvatar/MultipleUsers, Tag ❌

---

Guidelines:
- Don't add margin around the components.
- The bordered rectangle around each component in Figma is just for delimitation purposes.
- Put your component in `kibbeh/src/ui` and create a story in `kibbeh/src/stories`.
- Use tailwind for styling (see the tailwind config for theme info).
- You have to use TypeScript.
- Work on the `staging` branch.
- Please comment below if you start to work on a component, but only claim a component if you plan to work on it right away, so it doesn't block other people.
[...]

PR #2084, opened 2021-04-17
feat(kibbeh): ProfileTabs component w/ story

Mayowa-Ojo:
#2022

This PR:

- adds the ProfileTabs component

The profile tabs control which component is rendered below so I assume it will receive props that tell which tab is currently active from the parent component. Also, there will be a controller to update the active tabs in the parent component or wherever the state is managed when they're toggled.

TODO:
 - [ ]  Controller for the tabs

### DIFF
--- a/kibbeh/src/ui/ProfileTabs.tsx
+++ b/kibbeh/src/ui/ProfileTabs.tsx
@@ -1,0 +1,47 @@
+import React from "react";
+
+export interface ProfileTabsProps {
+   activeTab: string
+};
+
+export const ProfileTabs: React.FC<ProfileTabsProps> = ({ activeTab }) => {
+   return (
+      <div className="w-full flex items-center justify-around">
+         <button 
+            className={`py-1 text-primary-100 text-base font-bold hover:text-accent-hover 
+               ${activeTab == "about" && `border-b border-accent text-accent`}`
+            }
+         >
+            About
+         </button>
+         <button 
+            className={`py-1 text-primary-100 text-base font-bold hover:text-accent-hover 
+               ${activeTab == "rooms" && `border-b border-accent text-accent`}`
+            }
+         >
+            Rooms
+         </button>
+         <button 
+            className={`py-1 text-primary-100 text-base font-bold hover:text-accent-hover 
+               ${activeTab == "scheduled" && `border-b border-accent text-accent`}`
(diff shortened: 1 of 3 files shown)


kibbeh/src/ui/ProfileTabs.tsx
```diff
@@ -0,0 +1,47 @@
...
+export interface ProfileTabsProps {
+   activeTab: string
+};
+
+export const ProfileTabs: React.FC<ProfileTabsProps> = ({ activeTab }) => {
+   return (
+      <div className="w-full flex items-center justify-around">
+         <button 
+            className={`py-1 text-primary-100 text-base font-bold hover:text-accent-hover 
+               ${activeTab == "about" && `border-b border-accent text-accent`}`
```
ajmnz: Border should be `2px` wide (add it to `tailwind.config.js` otherwise you won't be able to put it).

kibbeh/src/ui/ProfileTabs.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export interface ProfileTabsProps {
+   activeTab: string
+};
+
+export const ProfileTabs: React.FC<ProfileTabsProps> = ({ activeTab }) => {
+   return (
+      <div className="w-full flex items-center justify-around">
+         <button 
+            className={`py-1 text-primary-100 text-base font-bold hover:text-accent-hover
```
ajmnz: Change hover effect to add border instead of changing the color.

```suggestion
            className={`py-1 text-primary-100 text-base font-bold border-b-2 border-primary-900 hover:border-accent transition
```

Mayowa-Ojo: Alright. Noted

Mayowa-Ojo: `border-primary-900` is a white border tho. Did you mean `border-accent`?

ajmnz: `primary-900` is the darkest gray we have. What I did here is match the border color with the background one (so it looks invisible) and then switch it to `accent` on hover.

Mayowa-Ojo: Ok my bad. Should be fine now

PR #2101, opened 2021-04-17
fix(button component): added tailwind focus rings to buttons

### LINKED ISSUE #2032: Change focused style of buttons
I really don't enjoy looking at the default button outline when in focus, I'd like to either remove it completely and add some other kind of visual clue that you actually clicked on it or restyle the outline. Right now it looks like this and I think it's pretty ugly: 

image

ghost:
Added tailwind styling to the 'Button.tsx' component that adds a better outline than a default one
(NOT TESTED)

fix #2032


ghost:
I couldn't run a local server on my machine because it kept giving errors so I couldn't test the changes, please do test them before accepting this PR, thanks

PR #2110, opened 2021-04-18
fix(auth redirect): user is now redirected to previously requested page after login

### LINKED ISSUE #2107: Login redirect to requested page isn't working
**Describe the bug**
The user isn't being redirected to the requested page after login. It's instead trying to redirect to /room/[id] which fails and returns the user to the dashboard.

**To Reproduce**
1. Copy a room link
2. Paste it into an incognito window / a browser where the user is signed out

**Expected behavior**
?next query parameter in the address bar should change to where the user originally requested

**Screenshots**
Screenshot 2021-04-17 at 23 32 32

**What device are you on?**
M1 Macbook Pro - Chrome

ollieleach:
fix #2107

This PR stops the "?next=" parameter in the address bar of the login screen being "/room/[id]" after trying to access a room while not being logged in and instead populates it with the correct requested page and redirects the user there after login.

I've tested on my machine with local accounts, but couldn't get logged in using OAuth - so please do test this using steps from #2107 before accepting this PR. 

Thanks!

### DIFF
--- a/kibbeh/src/modules/auth/useVerifyLoggedIn.ts
+++ b/kibbeh/src/modules/auth/useVerifyLoggedIn.ts
@@ -3,14 +3,19 @@ import { useEffect } from "react";
 import { useTokenStore } from "./useTokenStore";
 
 export const useVerifyLoggedIn = () => {
-  const { replace, pathname } = useRouter();
+  const { replace, query, pathname } = useRouter();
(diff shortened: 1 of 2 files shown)


samarmohan (approved):

benawad:
Breaks for `/u/[username]`

PR #2118, opened 2021-04-18
Added Basque and Tagalog back

rzsxdev:
- Added Basque back from #1349
- Added Tagalog back from #1417


dk-raw:
Make sure to put them in the language selector as well.

PR #2131, opened 2021-04-18
docs(readme.md): update README.md

MeztliRA:


benawad:
I'd rather just take kofta out of the readme

PR #2132, opened 2021-04-19
Fixed Title Visual Alignment

stupideveloper:
User Interface: Fixed the visual alignment on the feed title.

Don't understand why there was margin on left hand side, looked a bit off..


benawad:
it aligns with the text

PR #2142, opened 2021-04-19
Fixes #2137 - Hide empty room description & fix button outline

### LINKED ISSUE #2137: Kibbeh: Room description opens even though it's empty
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**What device are you on?**

**Additional context**

wes337:
If a room has no description, no button will be rendered.

I also fixed the strange outline around the button caused from margin/paddings.

Before:
before

After:
oivXriR2T6

Resolves #2137


rzsxdev:
One extra thing: I think the description text should have `break-words` Tailwind class or something because long links can go outside their container on smaller viewports.

dk-raw:
True

BronzW (approved):
Very epic

Issue #2171, opened 2021-04-19
Missing translations


Bqleine:
**Is your feature request related to a problem? Please describe.**
Some parts of the website are not translated (code-wise)

**Describe the solution you'd like**
Add the parts of the website to `translation.json` so they can be translated

**Additional context**
Here are the missing translations I could find :
![image](https://user-images.githubusercontent.com/51089082/115293120-39454600-a157-11eb-8351-a7831cd373b1.png)
*Dates in the explore rooms card*

![image](https://user-images.githubusercontent.com/51089082/115293196-50843380-a157-11eb-80b4-6465074c4121.png)
*The currently listening card ("listener", "leave")*

![image](https://user-images.githubusercontent.com/51089082/115293356-7f020e80-a157-11eb-9baf-5718db5480d2.png)
*Debug audio button*

![image](https://user-images.githubusercontent.com/51089082/115293419-93dea200-a157-11eb-9b37-a8d96a0da682.png)
*Sounds*

![image](https://user-images.githubusercontent.com/51089082/115293523-b96bab80-a157-11eb-9846-e8724598f0a8.png)
*App names in the [programmed rooms](https://dogehouse.tv/scheduled-rooms) reminder dropdown*

That's all I could find

BronzW:
I am going to create a pr that resolves this

Bqleine:
Great!

Bqleine:
Update there's this too :
![image](https://user-images.githubusercontent.com/51089082/115294635-203d9480-a159-11eb-999e-fd6862728c84.png)
*Delete scheduled room*

Bqleine:
Missing translations for the "with" in the room view
![image](https://user-images.githubusercontent.com/51089082/115378980-65051200-a1d1-11eb-80f8-d078d751def1.png)

PR #2173, opened 2021-04-19
[MERGE THIS BEFORE ANY OTHER TRANSLATION] Added a whole bunch of translation keys and created head titles for the pages that doesn't have them!

BronzW:
- Translation keys
- VSCode settings.json
- Header titles + translations for those
- Removed duplicate flex classnames


BronzW:
#2171

BronzW:
Did this right before I'm going to sleep. Don't expect response or fixes.

PR #2180, opened 2021-04-20
feat(kibbeh): In the EditProfileModal, make the bio textarea resizable

arjunmahishi:
The `textarea` is too small for the user's bio. It can only accommodate 2 lines by default. Making it resizable should be more convenient.

### DIFF
--- a/kibbeh/src/modules/user/EditProfileModal.tsx
+++ b/kibbeh/src/modules/user/EditProfileModal.tsx
@@ -131,7 +131,7 @@ export const EditProfileModal: React.FC<EditProfileModalProps> = ({
                 name="username"
               />
               <InputField
-                className={`mb-4`}
+                className={`mb-4 resizable-textarea`}
                 errorMsg={t("components.modals.editProfileModal.bioError")}
                 label={t("components.modals.editProfileModal.bioLabel")}
                 textarea
--- a/kibbeh/src/styles/globals.css
+++ b/kibbeh/src/styles/globals.css
@@ -106,7 +106,7 @@ html {
   overflow: auto;
 }
 
-textarea {
+textarea :not(.resizable-textarea) {
   resize: none;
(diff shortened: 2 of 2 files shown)


kibbeh/src/modules/user/EditProfileModal.tsx
```diff
@@ -131,7 +131,7 @@ export const EditProfileModal: React.FC<EditProfileModalProps> = ({
                 name="username"
               />
               <InputField
-                className={`mb-4`}
+                className={`mb-4 resizable-textarea`}
```
amitojsingh366, line 134: i don't think re-sizing is what we need since it's mostly associated with input. make it scroll instead?

amitojsingh366 (changes requested):

kibbeh/src/modules/user/EditProfileModal.tsx
amitojsingh366: also ask @goldyydev about this since i think he was the one that add the current limit

arjunmahishi: It's already scrollable. But you can never look at the whole bio at once.

amitojsingh366:
i say we should wait for @ajmnz and @goldyydev to hear what they have to say about this

ajmnz:
I agree that they should be able to fit the whole bio, but they shouldn't be resizable.

arjunmahishi:
> I agree that they should be able to fit the whole bio, but they shouldn't be resizable.

Ok. What's the alternative?
Scrolling is a pretty bad experience, especially on phone.

Is it a good idea to pre-size the textarea based on how much content is there? (They will still need to scroll while editing, if they write more stuff)

benawad:
I like auto-resizing better than default browser resizer

PR #2206, opened 2021-04-20
docs(DogeBook): scrolling through the docs page now works

Mvrs:
This PR:
- Allows users to scroll through the docs page
- This allows us to view our components definitions and types

Screenshot of Before: 
before-doc-fix

Now After (GIF):
docs-fix-after


Mvrs:
@overlisted 
something I felt like we really needed 😆

PR #2208, opened 2021-04-20
Fix Electron Scroll Bug

### LINKED ISSUE #2027: New Version Bug: Electron app's title bar makes it scroll weirdly
**Describe the bug**
The custom title bar makes the electron app scroll. I'm the one who added this so i know exactly what causes this. Adding the `margin-top` property on the UI does. *(see video below)*

**To Reproduce**
1. Go to 'next.dogehouse.tv' on the app
2. Scroll and see

**Expected behavior**
Shouldn't cause scrolling/should only scroll the room cards

**Screenshots**

https://user-images.githubusercontent.com/35400192/114820020-056cc800-9dc7-11eb-8c32-a5cafbeeb352.mp4

**What device are you on?**
Win 10, baklava: 1.0.55

**Additional context**
Need some css pro to solve this, the margin top property im talking about and showing in the video is found at `kibbeh/src/styles/electron-header.css`

MrDogeBro:
## Fixed
This is a simple change which fixes the electron scroll bug. 

## Description of what Changed
It overwrites the `.h-screen` class in the global css file. It sets the height of the class to the screen height minus the title bar height via the calc function. It uses the `--screen-height-reduction` variable as the amount to subtract which is `0px` by default. Then, in `_app.tsx`, if electron is present, the variable  `--screen-height-reduction` is set to `30px`, which is the height of the top bar. This then gets subtracted from the total view height because electron sees the bar as part of the view height. This essentially just tells it to set the view height to the total minus the bar if it is running in electron. 

If you would like any changes made or anything else like that, please tell me and I would be happy to do so. Thanks!

## Closes
- Closes #2027

### DIFF
--- a/kibbeh/src/pages/_app.tsx
+++ b/kibbeh/src/pages/_app.tsx
@@ -52,6 +52,10 @@ function App({ Component, pageProps }: AppProps) {
           useHostStore.getState().setData(platform);
         }
       );
+      document.documentElement.style.setProperty(
+        "--screen-height-reduction",
+        "30px"
+      );
     }
   }, []);
 
--- a/kibbeh/src/styles/globals.css
+++ b/kibbeh/src/styles/globals.css
@@ -35,6 +35,8 @@ p.bold {
   --color-accent: #fd4d4d;
   --color-accent-hover: #fd6868;
   --color-accent-disabled: #f5bfbf;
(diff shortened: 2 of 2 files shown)


amitojsingh366:
great job bro!

PR #2209, opened 2021-04-20
Remove the comment in commit-msg hook.

samarmohan:
There was a todo that said to do it when the new design hits, it did.


benawad:
later

PR #2220, opened 2021-04-21
kibbeh: add /download page

### LINKED ISSUE #1432: Add download link to website
A clear and concise description of what the problem is. Ex. I'm always frustrated when...
Users accessing the site may not know there is a desktop app

A download link somewhere on the site

**Describe alternatives you've considered**

**Additional context**
We would possibly need a way of identifying whether to not show the link if the user is already using the desktop app

### LINKED ISSUE #2187: Make the avatars more accessible to screen readers
Currently all avatars use an `alt` text of "avatar" which isn't screen reader friendly when having multiple avatars on screen (e.g. voice chat page)

Maybe make the `alt` text more "personalized". Like "Ben's avatar" or "John's avatar"

https://github.com/benawad/dogehouse/blob/611879da2bff5a172f880ac5ee0da82a383b08a7/kibbeh/src/ui/UserAvatar/SingleUser.tsx#L88-L90

### LINKED ISSUE #2212: Dropdown positioning breaks after scrolling the page
**Describe the bug**
After scrolling down on a page the positioning of dropdown menus is incorrect.

**To Reproduce**
1. Go to https://dogehouse.tv/scheduled-rooms
2. Scroll down so the top card is no longer visible
3. Click on the calendar button for the second card
4. Dropdown menu is positioned incorrectly

**Expected behavior**
The dropdown menu should be beside the calendar button that was clicked

**Screenshots**
Screen Shot 2021-04-21 at 11 44 26

**What device are you on?**
Google Chrome 89.0.4389.114
Macbook Pro 16-inch (2019) - macOS 10.15.7

**Additional context**
It seems related to using the `fixed` class in the `DropdownController`:
https://github.com/benawad/dogehouse/blob/staging/kibbeh/src/ui/DropdownController.tsx#L51

amitojsingh366:
- adds `/download` page to kibbeh *(closes #1432)*
- fixed #2187 *(it was one line change so didn't make separate pr)*
- ~~potentially fixes #2212 *(it was also a small fix)*~~

> added more than one thing in this pr as i don't know how to work on diff branches at the same time, ik its smtg to do with `git checkout` but not sure how to actually do it


ajmnz:
Why is `react-virtualized` removed from the lockfile?

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+function getOS() {
+  let isWindows = false;
+  let isMac = false;
+  let isLinux = false;
+  let isPhone = false;
+
+  const userAgent = window.navigator.userAgent;
+  const platform = window.navigator.platform;
+  const macosPlatforms = ["Macintosh", "MacIntel", "MacPPC", "Mac68K"];
```
dyoburon: nit: looks like window.navigator.platform can sometimes be "darwin" for macosPlatforms https://github.com/jsdom/jsdom/issues/1344

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const windowsPlatforms = ["Win32", "Win64", "Windows", "WinCE"];
+  const iosPlatforms = ["iPhone", "iPad", "iPod"];
+
+  let os = "Windows";
```
dyoburon, line 83: nit: Might be useful to set the os to "" instead of "Windows" in the case where a platform name doesn't exist in the arrays (either due to some being omitted or future changes). Feel free to ignore this nit if that is the desired behavior to have the default be Windows

amitojsingh366:
> Why is `react-virtualized` removed from the lockfile?

i have no idea... i just pulled and then added this, i didn't mess with lock

kibbeh/src/pages/download.tsx
amitojsingh366: i did set it to `""` before but typescript didnt like it. that's why i set it to windows now

kibbeh/src/pages/download.tsx
amitojsingh366: oh? lemme look more into this

amitojsingh366:
@ajmnz `react-virtualized` doesn't exist in the lockfile rn also?

image

overlisted:
it's GitHub not Github

ajmnz:
Resolve conflicts

amitojsingh366:
> Resolve conflicts

ok my bad

ofsho:
test

kibbeh/src/pages/download.tsx
amitojsingh366: added this

kibbeh/src/global-stores/useScrollStore.ts
```diff
@@ -0,0 +1,16 @@
+import create from "zustand";
+import { combine } from "zustand/middleware";
+
+export const useScrollStore = create(
```
benawad: what is this about? Remove console.logs related to this (you should be splitting the prs so I can merge the ones that don't need changes)

benawad (changes requested):

kibbeh/src/global-stores/useScrollStore.ts
amitojsingh366: this was to fix #2212, but it doesn't work as i expected... ill just remove/comment the code for it then

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,102 @@
...
+  } else if (iosPlatforms.indexOf(platform) !== -1) {
+    os = "iOS";
+    isPhone = true;
+  } else if (windowsPlatforms.indexOf(platform) !== -1) {
+    os = "Windows";
+    isWindows = true;
+  } else if (/Android/.test(userAgent)) {
+    os = "Android";
+    isPhone = true;
+  } else if (!os && /Linux/.test(platform)) {
```
dzmitry-vasileuski: Default value of `os` variable is `Windows`. So you will never pass `!os` condition here. Let's remove it.

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,102 @@
...
+  let isPhone = false;
+
+  const userAgent = window.navigator.userAgent;
+  const platform = window.navigator.platform;
+  const macosPlatforms = ["Macintosh", "MacIntel", "MacPPC", "Mac68K", "darwin"];
+  const windowsPlatforms = ["Win32", "Win64", "Windows", "WinCE"];
+  const iosPlatforms = ["iPhone", "iPad", "iPod"];
+
+  let os = "Windows";
+  if (macosPlatforms.indexOf(platform) !== -1) {
```
dzmitry-vasileuski, line 84: I believe it would be great to replace such constructions with `includes`
```
if (macosPlatforms.includes(platform)) {
...
}
```

kibbeh/src/modules/scheduled-rooms/AddToCalendar.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { ReactNode, useEffect, useMemo, useRef, useState } from "react";
+// import { useScrollStore } from "../../global-stores/useScrollStore";
```
dzmitry-vasileuski: It's bad practice to keep commented-out code in VCS. Just remove all this stuff.

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,102 @@
...
+          >
+            {t("pages.download.visit_gh")}
+          </Button>
+        ) : null}
+      </div>
+    </>
+  );
+}
+
+function getOS() {
```
dzmitry-vasileuski, line 71: Would it be better to move all this code to a separate module so it can be reused in other parts of the app? Also would be great to have constants for all these sting values.

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,102 @@
...
+      let res = false;
+      var xmlHttp = new XMLHttpRequest();
+      xmlHttp.open(
+        "GET",
+        "https://api.github.com/repos/benawad/dogehouse/releases/latest"
+      );
+      xmlHttp.send(null);
+      xmlHttp.onreadystatechange = () => {
+        if (xmlHttp.responseText && !res) {
+          res = true;
```
dzmitry-vasileuski, line 30: What is the reason to use this trick with `res`? I believe you can just do not pass the empty array as the second parameter of `useEffect` so it's triggered only once instead of every rendering.

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,102 @@
...
+          let data = JSON.parse(xmlHttp.responseText);
+          let tag = data.tag_name;
+          if (tag) {
+            let v = tag.replace("v", "");
```
dzmitry-vasileuski: I would recommend replacing `v` variable name with the full word `version` (related to placeholder in links as well)

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,102 @@
...
+      xmlHttp.open(
+        "GET",
+        "https://api.github.com/repos/benawad/dogehouse/releases/latest"
+      );
+      xmlHttp.send(null);
+      xmlHttp.onreadystatechange = () => {
+        if (xmlHttp.responseText && !res) {
+          res = true;
+          let data = JSON.parse(xmlHttp.responseText);
+          let tag = data.tag_name;
```
dzmitry-vasileuski: `const` will be more appropriate for both `data` and `tag` variables.

kibbeh/src/pages/download.tsx
```diff
@@ -0,0 +1,102 @@
...
+  Linux:
+    "https://github.com/benawad/dogehouse/releases/download/{{tag}}/DogeHouse-{{v}}.AppImage",
+};
+
+export default function Download() {
+  const [downloadFailed, setDownloadFailed] = useState(false);
+  const { t } = useTypeSafeTranslation();
+  useEffect(() => {
+    let os = getOS();
+    if (!os.isPhone) {
```
dzmitry-vasileuski, line 20: Use early return to decrease nesting level.

```
if (os.isPhone) {
  setDownloadFailed(true);
  return;
}

... your code goes here ...

```

kibbeh/src/pages/download.tsx
amitojsingh366: yes, sure!

kibbeh/src/pages/download.tsx
amitojsingh366: 👍

kibbeh/src/pages/download.tsx
amitojsingh366: this shouldn't make much of a difference

kibbeh/src/modules/scheduled-rooms/AddToCalendar.tsx
amitojsingh366: it might be used later but ok...

kibbeh/src/pages/download.tsx
amitojsingh366: for now this isn't used anywhere else, so for now let's just keep it here...

kibbeh/src/pages/download.tsx
amitojsingh366: i don't see the urgency in this but i changed it...

kibbeh/src/pages/download.tsx
amitojsingh366: yep sure will change it

benawad:
rip out all the `useScrollStore` stuff (we are not going to solve that bug in that way), fix conflict, and then I'll merge

amitojsingh366:
> 
> 
> rip out all the `useScrollStore` stuff (we are not going to solve that bug in that way), fix conflict, and then I'll merge

give me 2 seconds

amitojsingh366:
ok @benawad it's ready now

vi241:
cool

PR #2258, opened 2021-04-22
Make it easier to add environment variables

samarmohan:
Create two files (one for windows, one for unix-based) with all the environment variables needed and update CONTRIBUTING.md accordingly.


samarmohan:
Why are all my previous commits appearing?

PR #2260, opened 2021-04-22
feat(kousa,docs): make it easier to set environment variables by adding scripts.

samarmohan:
Create two files (one for windows, one for unix-based) with all the environment variables needed and
update CONTRIBUTING.md accordingly.


CONTRIBUTING.md
```diff
@@ -123,25 +123,7 @@ $ CREATE DATABASE kousa_repo2;
...
-export TWITTER_BEARER_TOKEN=
-export GITHUB_CLIENT_SECRET=
-export SENTRY_DNS=
-export API_URL=http://localhost:4001
-export WEB_URL=http://localhost:3000
-export PORT=4001
-```
-
-> You can save these variables in a `.txt` and run `source path/to/file.txt`
+Navigate to `/kousa` and run `call env.windows.bat` if you are on Windows or `source env.unix.sh` if you are on a Mac or Unix based system. This will set the necessary environment variables.
```
amitojsingh366, line 126: this is not what we want, the env variables shown here are just examples and need to be filled... we don't want to supply the contributors with our own versions of them.

amitojsingh366 (changes requested):

CONTRIBUTING.md
samarmohan: What should I do then?

amitojsingh366: leave it as is....

samarmohan:
Anything stopping this from being merged

samarmohan:
sorry that was a test using GH cli

PR #2269, opened 2021-04-23
too dumb emote

Mudrank:


dk-raw:
We are not accepting new emotes rn

PR #2293, opened 2021-04-24
CROATIAN | Added new translations, fixed some old ones 

tomicleon:
Please, whoever keeps changing "pratitelji" to "slijedbenici" in "viewUser", stop, it's a completely incorrect translation in the context of social media. Also whoever keeps changing "following" to "sljedeći", just know it means "next" and not "following", the correct translation is "pratiš" or "pratite".


ajmnz:
Resolve conflicts and we merge

PR #2306, opened 2021-04-25
Match Figma Design (User Profile)

rzsxdev:
Corrected some styles for the User Profile page.

Changed the `p` tag to a `div` tag because the global `p` tag styles font-weight is 500 and almost all elements that have 500 font-weight on Figma are 400 on the actual site.


ajmnz:
I'll close this as I'm currently changing all p tags to not use that font weight

PR #2320, opened 2021-04-25
don't do this again!

dark-elixir-shibe:
@benawad high priority; CI tests are broken until this is resolved.


nadirabbas:
dont do what?

gallexme:
> dont do what?

editing a existing Migration, instead of creating a new Migration, it breaks the whole migration chain and u may have a inconsistent database(a system which already applied the migration, will not again apply it)

PR #2323, opened 2021-04-25
baklava: fix global keybinds

amitojsingh366:
- fixes global keybinds for baklava. *(was calling hook from a non-functional component)*


overlisted (approved):
looks good to me

PR #2333, opened 2021-04-26
Added translations to CONTRIBUTING.md

HonbraDev:
@TheOtterlord


TheOtterlord (approved):
Thanks.
Looks good, though it makes me wonder if we should set up json validation for the locales 🤔 
Anyway that's another issue

PR #2355, opened 2021-04-26
chore(Kibbeh): add raw text to translations

lbarthon:
## Changes
- Repairs the `he` translation file that had a weird trailing character and a missing dquote
- Adds 2 new translation keys:
  <details><summary>pages.viewUser.about</summary></details>
  <details><summary>components.modals.editProfileModal.bannerUrlLabel</summary></details>
- Translate some stuff in french / `fr`


rzsxdev:
The word "with" is a bit problematic for translations because some langauges don't have a word for it and instead use suffixes after the nouns.

lbarthon:
> The word "with" is a bit problematic for translations because some langauges don't have a word for it and instead use suffixes after the nouns.

Oh, should I add the name as a placeholder in the translation ?
Edit: Cannot be done because usernames are in bold, I just removed the commit

PR #2361, opened 2021-04-26
Fix #2244 - Overflowing leave button on small screen

### LINKED ISSUE #2244: Buttons overflowing in small screens
**Describe the bug**

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the chat ui in smaller phones such as moto g4

**Expected behavior**
The ui should adapt to the small screen.  

**Screenshots**
Screen short of the overflowing ui
Screenshot_20210422_140540

**What device are you on?**
moto g4

**Additional context**
The bottom buttons on the chat page overflows when there is a smaller screen.

wes337:
Uses flex wrap, and justifies content on XS screens only. Only affects the panel with many options. Resolves #2244 

Before:
before

After:
after

Here's a GIF:
cYf8tbzH21


wes337:
Not sure why the tests are failing. They are all passing. Also is there now more rules to how commit messages should be?

benawad:
If you pull and push again the gh action should pass when the tests should pass

for commit messages use conventional commit

nadirabbas:
I think the `leave` button should have full width in column flow. What do you think @ajmnz ?

ajmnz:
> I think the `leave` button should have full width in column flow. What do you think @ajmnz ?

I'm not really a fan of this, but guess everything in the mobile version is temp. So yeah, let's go full width

wes337:
Changes done. Also the chat button was a different size after the latest pull here.
XS:
xs

Small:
sm

Md:
md

wes337:
Seems like the pipline is still failing, for ESLint reasons on files unrelated to this PR, even after pulling and merging the latest Staging branch.

Also my commit messages are still not passing. I am adding the "type" to the message (`fix: something`). What am I doing wrong?

kibbeh/src/shared-hooks/useSetDeaf.ts
```diff
@@ -2,14 +2,14 @@ import { Wrapper } from "@dogehouse/kebab";
 import { useDeafStore } from "../global-stores/useDeafStore";
 import { useWrappedConn } from "./useConn";
 
+export const setDeaf = (conn: Wrapper, value: boolean) => {
```
wes337: This is entirely unrelated, but was causing the `lint` step to fail in the GH actions.

benawad:
@ajmnz I agree, its going to be very rare someone has a phone this size so I'm fine with this

kibbeh/src/shared-hooks/useSetDeaf.ts
amitojsingh366: pull latest and push, should be fixed

amitojsingh366:
use proper commit messages please *(`yarn commit` or `yarn commit:signed`)*

benawad:
> Also my commit messages are still not passing. I am adding the "type" to the message (`fix: something`). What am I doing wrong?

The thing in parens should be one of the scopes, in this case `kibbeh` and you need to add a scope to that one commit
image

benawad:
if you pull, I think the eslint error should go away

wes337:
Even though I used yarn commit, it's still not happy about my previous commits.

amitojsingh366:
> 
> 
> Even though I used yarn commit, it's still not happy about my previous commits.

one of your commits uses `fix(lint)`, it won't accept that. see the screenshot below it contains all the scopes accepted:
image

PR #2373, opened 2021-04-27
Fix auto mentioning names without @ symbols

TheBlueOompaLoompa:


Gers2017:
Please give this man a medal for this beautiful code

kibbeh/src/lib/createChatMessage.ts
```diff
@@ -18,7 +18,7 @@ export const createChatMessage = (
   const testAndPushToken = (item: string) => {
     const isLink = linkRegex.test(item);
     const withoutAt = item.replace(/@|#/g, "");
-    const isMention = roomUsers.find((m) => withoutAt === m.username);
+    const isMention = roomUsers.find((m) => withoutAt === m.username && item.includes(`@${withoutAt}`));
```
lbarthon, line 21: The replace line seem to be handling both `@` and `#`, and your includes only looks for an `@`

TheBlueOompaLoompa: My code only looks for an @ symbol because it only affects when you try to mention someone and not when you whisper to someone. Whisper uses #@. Also the issue is related to when you use a word that happens to be a username.

lbarthon: If both features are still working, it's ok

TheBlueOompaLoompa: image
Just tested it and it works perfectly!

amitojsingh366:
this isn't how we want to do this unfortunately...

PR #2377, opened 2021-04-27
fix width of chat on wider layouts

TheBlueOompaLoompa:
image
The chat was set to 325px width I think, so I changed it to 24.5vw to accommodate people with larger screen sizes.


ajmnz:
We’re not changing this at the moment.

PR #2388, opened 2021-04-27
Add /bot/auth wrapper to kebab

overlisted:
I'm going to make at least 2 more PRs improving kebab soon.


amitojsingh366:
do we need this in kebab?

BronzW:
Why not?

amitojsingh366:
> 
> 
> Why not?

kebab should be a socket only thing i think, the users should get their bot tokens themselves i think

PR #2392, opened 2021-04-27
ci(kibbeh): add scripts to run Cypress headlessly

mateuszkoncikowski:
1. Increasing global timeout (running on the remote server in headless mode will be considerably slower). We could add configuration to differentiate the execution environment but this could be done later. 
2. Running tests in Chrome browser instead of the default one (Electron based). The frontend contains Electron specific logic which interferes with tests in headless mode thus the change.

### DIFF
--- a/.github/workflows/label-pr.yml
+++ b/.github/workflows/label-pr.yml
@@ -1,6 +1,7 @@
 name: Auto label pr
 
-on: [pull_request]
+on:
+  - pull_request_target
 
 jobs:
   triage:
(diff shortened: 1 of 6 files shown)


A-Yamout:
Resource not accessible by integration?

amitojsingh366:
> 
> 
> Resource not accessible by integration?

that's not a problem, im fixing it

amitojsingh366:
@mateuszkoncikowski can you pull latest and push

mateuszkoncikowski:
Updated.

PR #2394, opened 2021-04-27
feat(pilaf): Add StackNavigationProps interface for navigation and route

dwieners:
This PR adds a suggestion for dealing with navigation props and route props. In future versions this should also be adapted to the hook `useNavigation`. The desired behavior can be achieved as follows: 
```ts
const navigation = useNavigation<StackNavigationProps<RootStackParamList, "Room">>()
```


goldyydev:
Do you mind renaming the commits from feature(... to feat(...?

Issue #2356, opened 2021-04-26
Chat does not resize properly when changing screen resolutions


Temi-pinheiro:
**Describe the bug**
When in a room, if you are on a laptop/desktop and your window size is mobile or tablet, when you click into chat and send or receive a couple of messages and then later resize your window back to desktop, there is large spacing between the messages.

**To Reproduce**
1. Resize your window to a tablet or mobile size
2. Click on the chat box
3. Send or receive a couple of messages
4. Resize the window to desktop

**Expected behavior**
It's supposed to look fine and resize on refresh, my guess is its a calculation issue affecting padding or margin. If you type any message after resizing, everything goes back to normal

**Screenshots**
![image](https://user-images.githubusercontent.com/63761067/116126303-026eb300-a6b6-11eb-8e75-c4aa87ab8464.png)
![image](https://user-images.githubusercontent.com/63761067/116126405-23370880-a6b6-11eb-9e1a-1dcee1759f88.png)
![image](https://user-images.githubusercontent.com/63761067/116126472-34801500-a6b6-11eb-9d5b-50d9abb31df5.png)

**What device are you on?**
HP Pavillion 15 inches.

nadirabbas:
what's not proper here?

Gers2017:
I'm working on this

lee1409:
Is anyone still working on this? If not, I'm willing to help working on this issue. 😁

nadirabbas:
I believe nobody's working on it

Gers2017:
yeah I'm nobody thanks 😄

nadirabbas:
@Gers2017 sorry I thought you abandoned it.

Gers2017:
I did. But as my uncle said don't give on your dreams

PR #2395, opened 2021-04-27
fix(kibbeh): removed inline styles from RoomChatList

Gers2017:
Removed Inline styles that used position absolute and translate

#2356

ex2
ex3
ex1


benawad:
those are needed for virtualization, I'm guessing this change would not work when the list fills up and you start scrolling

Issue #2376, opened 2021-04-27
Language flag/ISO code does not show


BronzW:
![image](https://user-images.githubusercontent.com/50546858/116197141-a008c780-a734-11eb-94a7-a5153306b1d9.png)

amitojsingh366:
i think this is intentional

BronzW:
Why

BronzW:
@ajmnz Changed it in his latest PR so the flags is their own property in LanguageSelector.tsx but he didn't add support for it in the html.

amitojsingh366:
im not sure about this, ill ask @ajmnz to look into this

ajmnz:
Many times flags don’t respresent all the people that speak a certain language. It can even offend people in some occasions and that’s why I removed them. 

I left them as a separate property because I felt that they could be useful in the future, but the change was intentional.

PR #2415, opened 2021-04-28
Fixes #2136 Language sorting and labels

### LINKED ISSUE #2136: Language sorting and labels
**Describe the bug**
The languages are sorted by their language code. The labels are translated. This makes it difficult to navigate through the list and find your language. The label shouldn't have the emoji flag inside of it _(this could be another object key)_.

Open the list of languages in the language selector and try to find any language other than English.

**Screenshots**
image

wes337:
Language labels are now translated and sorted A-Z. Novelty languages are put to the bottom. Uses the flag in the SettingsItem icon prop.

Before (hard to find my own language):
beforeCapture
 Finnish is suomi, but because we sort by code it's near French.

After:
after

Languages will need to be translated:
after2

Fixes #2136


BronzW:
Why do you need to add translation keys to language names?

wes337:
> 
> 
> Why do you need to add translation keys to language names?

It's really hard to find your own language in the current way the languages are sorted. See my example with Finnish.

BronzW:
Yes but why did you add translation keys.

BronzW:
To translation.json

wes337:
> 
> 
> To translation.json

Is there another way to translate these strings? How else can you know how to say "Latvian" in French?

We can't just sort them by the ISO code (for example "fi" for Finnish, but the language label is Suomi) if we're just using each language label in its own language. If you look at the language selector right now, it would appear they're in random order. It's not clear what the ISO code is if we're just sorting it by label.

rzsxdev:
Another potential sorting system is the system for example Wikipedia uses. The languages are in their native names and sorted alphabetically but are also grouped by their writing system: Latin script first, then Cyrillic, Abjads (Arabic-like scripts), Abuguidas (Hindi-like scripts), and then the Chinese-like scripts.

wes337:
> 
> 
> Another potential sorting system is the system for example Wikipedia uses. The languages are in their native names and sorted alphabetically but are also grouped by their writing system: Latin script first, then Cyrillic, Abjads (Arabic-like scripts), Abuguidas (Hindi-like scripts), and then the Chinese-like scripts.

Where? I only see Wikipedia sorting languages based on articles written.

rzsxdev:
At least on the mobile browser version you can access the list of languages by clicking the icon on the top left on any article page:

ajmnz:
Why did you add the flags? https://github.com/benawad/dogehouse/issues/2376#issuecomment-827418623

ajmnz:
> Another potential sorting system is the system for example Wikipedia uses. The languages are in their native names and sorted alphabetically but are also grouped by their writing system: Latin script first, then Cyrillic, Abjads (Arabic-like scripts), Abuguidas (Hindi-like scripts), and then the Chinese-like scripts.

I agree, each language should be in its native name. Makes no sense to have English translated to French, if you know a language enough to use it in the platform, you'll know how it's called natively.

amitojsingh366:
could you please make a new pr in which only the sorting alphabetically part is done *(it will be better if it sorts dynamically)*, we discussed about this and don't need the other things in this pr

Issue #2397, opened 2021-04-27
Bots shouldn't be included in member counts


overlisted:
Rooms like this one are at the very top of the room list, because bot accounts are included in the member count:
![image](https://user-images.githubusercontent.com/66535446/116319942-a3916280-a7b7-11eb-9b91-caf31876791e.png)

I don't think that it should be like that, and the room on that screenshot should be as low on the list as a room with just 3 members.

overlisted:
Proof that it's at the top:
![image](https://user-images.githubusercontent.com/66535446/116320675-f3bcf480-a7b8-11eb-87c4-8b26cf5bbbb2.png)

nadirabbas:
why not?

overlisted:
They can be abused to gain more members and get higher on the list

HoloPanio:
I both agree and disagree.  I think bot accounts should add to the number of accounts in the room, but that is all it does, it doesn't push you up.  So if there is a room with 5 bots and 10 users, these values should be separate and you sift through the rooms using the number of users, but you add that number of bots just on the UI.

Mayowa-Ojo:
I think there should be a limit to the number of bots per room

nadirabbas:
@HoloPanio yeah, we should exclude bot accounts when sorting

PR #2434, opened 2021-04-29
Solving issue #2397 - issues with bot accounts being counted for room size

dqnk:
- add support for tracking bots in elixir database
- added counting of people and bots in rooms
- add fields for bot variables in the database
- add environment variables file in /kousa to .gitignore

I tested the PR and it shouldn't break anything, but I was unable to create bot accounts on my localhost to test if they actually don't count to final room count anymore #2397.


kousa/lib/beef/schemas/user.ex
```diff
@@ -77,6 +78,7 @@ defmodule Beef.Schemas.User do
     field(:avatarUrl, :string)
     field(:bannerUrl, :string)
     field(:bio, :string, default: "")
+    field(:isBot, :boolean, default: false)
```
amitojsingh366: we don't want this, we already have `botOwnerId` and we can track if a user is a bot or not by using that

amitojsingh366 (changes requested):

kousa/lib/beef/schemas/user.ex
dqnk: thanks for noticing, will fix

dqnk: So if I understand correctly, botOwnerId (in `kousa/lib/beef/schemas/user.ex`) is either `null/nil` for regular accounts or an `Ecto` type for bots?

amitojsingh366: also we probably don't wanna do this at the backend, this can maybe be done fully on the client side

amitojsingh366:
we probably don't wanna do this at the backend and maybe do it client side only but that's upto @benawad

amitojsingh366:
also please use proper commit messages *(`yarn commit` or `yarn commit:signed`)*

amitojsingh366:
and your elixir tests are failing

benawad:
idk if I want this behavior, I will think on it

dqnk:
Note that my last commit probably doesn't solve all the issues with elixir.

amitojsingh366:
> idk if I want this behavior, I will think on it

so do we close this then

benawad:
I'm going to close this for now because we are going to be moving the # of users count out of the database

PR #2438, opened 2021-04-29
fix(baklava): fix dumb default deafen keybind

### LINKED ISSUE #2435: Electron app deafen (Ctrl-X) keyboard shortcut applies even when the app is minimized and another window is in focus
**Describe the bug**
The Electron Desktop Application's keyboard shortcut for deafening a room is triggered even when the app is minimized and another window is active. For example, I could be in VScode trying to cut out some code with DogeHouse minimized but still open, but instead of cutting the code out, it deafens the audio in the room even though DogeHouse is minimize.

**To Reproduce**
1. Open the Electron desktop app
2. Join any room
3. Hit minimize in the top right, and open any other window (and focus on it)
4. Press Ctrl-X.
5. The room will deafen, but the Ctrl-X shortcut should apply only when DogeHouse is in focus

**Expected behavior**
Ctrl-X does whatever it's mapped to do for the current window. For example, if a code editor is open and focused on and the Electron app is minimized while listening in a room, Ctrl-X should cut out highlighted text (or do whatever it's programmed to do) instead of deafening the room.

**What device are you on?**
Windows 10

amitojsingh366:
- changed `Control+x` to `Control+1`
- fixes #2435

### DIFF
--- a/baklava/src/utils/keybinds/index.ts
+++ b/baklava/src/utils/keybinds/index.ts
@@ -25,7 +25,7 @@ import { stopRPC } from "../rpc";
 export let CURRENT_REQUEST_TO_SPEAK_KEY = "Control+8";
 export let CURRENT_INVITE_KEY = "Control+7";
 export let CURRENT_MUTE_KEY = "Control+m";
-export let CURRENT_DEAF_KEY = "Control+x";
+export let CURRENT_DEAF_KEY = "Control+1";
 export let CURRENT_CHAT_KEY = "Control+9";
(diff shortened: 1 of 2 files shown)


yiays:
Ctrl+1 clashes with the keyboard shortcut to switch to tab 1 in Chrome. In my testing, both happen at once.

ghost:
It does the same on Firefox.

yiays:
Discord uses `Ctrl+Shift+M` for mute and `Ctrl+Shift+D` for deafen, perhaps we should take inspiration from this?

PR #2451, opened 2021-04-30
Switch everything to lf

samarmohan:
We should be using LF everywhere, but that wasn't the case. I found a script that solved this problem.


amitojsingh366:
improper commit messages, also this pr is very big and we can't possibly review it. you could have snuck in some malicious code that we have no way to review.

PR #2453, opened 2021-04-30
style(global): Fix eslint line ending rules

samarmohan:
The linebreak-style option is now `off` instead of `error`.


goldyydev:
I think that a better approach to this would be to make a git hook to convert all __edited__ files to LF (maybe stylelint does this, please check it).

benawad:
yeah im pretty sure we want this

PR #2477, opened 2021-05-01
Random polishments to kebab

overlisted:
I will rewrite the wrapper very soon!


overlisted:
lmao i should really make git sign my commits by default

kebab/src/websocket/raw.ts
```diff
@@ -1,26 +1,26 @@
...
 
 export type Token = string;
-export type FetchID = UUID;
+export type Ref = UUID;
 export type Opcode = string;
 export type Logger = (
   direction: "in" | "out",
   opcode: Opcode,
   data?: unknown,
-  fetchId?: FetchID,
```
benawad: you need to keep fetchId until all the new casts/calls have been put in, only the ones with keys like `room:ban`  work with ref

benawad (changes requested):

kebab/src/websocket/raw.ts
overlisted: oh

PR #2486, opened 2021-05-02
feat(kibbeh): add speaker badge near to username in chat

### LINKED ISSUE #1413: Speaker chat badges
Show a `📣` next to your username in chat, to show that you are a speaker, if you are.

You could also add other things like `👑` for room owner, `🔨` for moderator, and such.

Speaker Chat Badge

junedkhatri31:
fix #1413

**THIS IS INPROGRESS**

Need to performance test once with at least hundreds of users and thousands of messages.


junedkhatri31:
Can anyone help me what would be the best way to performance test this?
I am concerned about this code
```js
const user = users.find((u) => u.id === m.userId);
```

junedkhatri31:
Screenshot 2021-05-02 at 2 07 30 PM
This is how it looks like

ajmnz:
> Screenshot 2021-05-02 at 2 07 30 PM
> 
> This is how it looks like

Add the emoji before the username and use twemoji instead of the Apple ones. (maybe @goldyydev can point you in the right direction on that)

junedkhatri31:
Updated.
Screenshot 2021-05-02 at 5 31 21 PM

Changed css in global bcoz emoji was reflecting display block which consume width 100%. `emoji` css is only used in Settings label. I've checked, nothing breaks there.

kibbeh/src/modules/room/chat/RoomChatList.tsx
```diff
@@ -43,6 +45,16 @@ export const RoomChatList: React.FC<ChatListProps> = ({ room }) => {
...
   });
 
+  const getBadgeIcon = (m: Message) => {
+    const user = users.find((u) => u.id === m.userId);
+    const isSpeaker = room.creatorId === user?.id || user?.roomPermissions?.isSpeaker;
+    let emoji = null;
+    if (isSpeaker) {
+      emoji = "📣";
+    }
+    return emoji && <Twemoji text={emoji} style={{ marginRight: "1ch" }}/>;
```
benawad: can you add a title to it so people can see what it means on hover

benawad (changes requested):

kibbeh/src/modules/room/chat/RoomChatList.tsx
junedkhatri31: Done

PR #2501, opened 2021-05-02
Feature/numactive

nadirabbas:
people active in last two days. like these people are considered "regular users"

### DIFF
--- a/kousa/lib/broth/routes/stats.ex
+++ b/kousa/lib/broth/routes/stats.ex
@@ -17,7 +17,7 @@ defmodule Broth.Routes.Stats do
     two_days_ago = Timex.now |> Timex.shift(days: -2)
     query =
       from u in User,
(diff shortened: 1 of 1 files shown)


kousa/lib/broth/routes/stats.ex
```diff
@@ -12,13 +14,20 @@ defmodule Broth.Routes.Stats do
   alias Beef.Schemas.User
 
   defp getStats do
-    d = {DateTime.now!("Etc/UTC"), {Repo.aggregate(User, :count, :id)}}
+    two_days_ago = Timex.now |> Timex.shift(days: -2)
+    query =
+      from u in User,
+        where: u.lastOnline > ^two_days_ago;
```
benawad: or if they are currently online

benawad (changes requested):

Issue #2483, opened 2021-05-01
follow user button is not responsive in mobile view


munsif12:
![IMG_20210502_030252](https://user-images.githubusercontent.com/53260498/116796041-ed00eb00-aaf2-11eb-8027-1b4bfcd3db8f.jpg)
![IMG_20210502_030450](https://user-images.githubusercontent.com/53260498/116796070-2d606900-aaf3-11eb-85b5-b43ed9fd1a8c.jpg)

junedkhatri31:
What's the screen size of your mobile?

munsif12:
> What's the screen size of your mobile?
@junedkhatri31 
It's 720 × 1544 pixels
And mobile is vivo y15

amitojsingh366:
can you try a different browser?

Alirezamxz:
Checked in 3 different browsers, i think it's ok
![bn](https://user-images.githubusercontent.com/47672255/116853177-dbab0200-ac0a-11eb-8687-f73ceb74eadb.jpg)

amitojsingh366:
ok then since this issue is fixed ill close it

munsif12:
> Checked in 3 different browsers, i think it's ok
> ![bn](https://user-images.githubusercontent.com/47672255/116853177-dbab0200-ac0a-11eb-8687-f73ceb74eadb.jpg)

@Alirezamxz Thanks brother

munsif12:
@amitojsingh366 
@Alirezamxz 
@junedkhatri31 
Guys i am soory but i tested  the same issue again on 3 different browers  ( Google Chrome ,opera mini and firefox )but its still not working
I am attaching screenshot with browers name on them 
hoping the best 👍
![Screenshot_20210503_231543](https://user-images.githubusercontent.com/53260498/116916762-3114fc80-ac67-11eb-884b-1dd2a9aeb851.jpg)
![IMG_20210503_232135](https://user-images.githubusercontent.com/53260498/116916790-396d3780-ac67-11eb-962e-b398dcea7441.jpg)
![Screenshot_20210503_232007](https://user-images.githubusercontent.com/53260498/116916840-468a2680-ac67-11eb-852f-04f14143235d.jpg)
![Screenshot_20210503_231401](https://user-images.githubusercontent.com/53260498/116916881-543fac00-ac67-11eb-8783-8e4b19eff2b0.jpg)

amitojsingh366:
oh this is still an issue? ok ill reopen this

[...]

PR #2511, opened 2021-05-03
fix(kibbeh): Fix responsive follow user button in mobile view

jhony-v:
Adding a follow user button below the  user profile image and description in mobile view #2483.

### DIFF
--- a/kibbeh/src/ui/ProfileHeader.tsx
+++ b/kibbeh/src/ui/ProfileHeader.tsx
@@ -72,118 +72,122 @@ export const ProfileHeader: React.FC<ProfileHeaderProps> = ({
           }
         }}
       />
-      <div className="flex mr-4 ">
-        <SingleUser
-          isOnline={user.online}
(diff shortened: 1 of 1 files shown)


kibbeh/src/ui/ProfileHeader.tsx
```diff
@@ -72,118 +72,122 @@ export const ProfileHeader: React.FC<ProfileHeaderProps> = ({
...
+            ) : (
+              ""
+            )}
+            {canDM ? (
+              <Button size="small" color="secondary" icon={<SolidMessages />}>
+                Send DM
+              </Button>
+            ) : (
+              ""
+            )}
```
Tchoukoualeu, line 189:
```suggestion
            {canDM && (
              <Button size="small" color="secondary" icon={<SolidMessages />}>
                Send DM
              </Button>
            )}
```

Tchoukoualeu: And this goes for all other cases.

amitojsingh366:
this issue seems to be fixed now and hence there is no need for this pr

Issue #2485, opened 2021-05-02
Mobile UX Show more follow button issue


s3ansh33p:
**Describe the bug**
On mobile, whilst in a room and viewing a user's followers, the room overlay is placed over the show more button. This makes it hard to click the show more button instead of the overlay.

**To Reproduce**
1. (whilst in a room) Go to my profile - https://dogehouse.tv/u/admin
2. Click on followers
3. Scroll down to the show more button

**Expected behavior**
The room overlay should be above the show more button.

**Screenshots**
https://ibb.co/NLPtNSm

**What device are you on?**
iPhone XS, Chrome

**Additional context**
N/A

ajmnz:
I think it could be solved either by adding a bigger margin at the bottom of the list, or by changing the button to infinite scroll

wes337:
@ajmnz Maybe even both? Because you could get to the bottom of the infinite scroll and the room badge would still be on top of the last user.

![image](https://user-images.githubusercontent.com/9464690/116811793-e5bdf980-ab53-11eb-9511-ce32a726a043.png)

Also somewhat related --- having a long name will cause the "Follow" button to be pushed off screen.

ajmnz:
> @ajmnz Maybe even both? Because you could get to the bottom of the infinite scroll and the room badge would still be on top of the last user.

Yep, makes sense.


> Also somewhat related --- having a long name will cause the "Follow" button to be pushed off screen.

Mobile version is kinda temp, we're currently refactoring it completely to match Figma. So yeah, many things need to be polished

PR #2514, opened 2021-05-03
fix(kibbeh): fix intersection observer hook

wes337:
The intersection observer hook I created in #2491 was only getting set on mount. This fixes that and the entry should be updating correctly now.

Relates to #2485

### DIFF
--- a/kibbeh/src/shared-hooks/useIntersectionObserver.ts
+++ b/kibbeh/src/shared-hooks/useIntersectionObserver.ts
@@ -10,11 +10,11 @@ export const useIntersectionObserver = ({
   const observer = useRef<IntersectionObserver>(null);
 
   useEffect(() => {
-    if (node) {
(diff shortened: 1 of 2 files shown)


kibbeh/src/shared-hooks/useIntersectionObserver.ts
```diff
@@ -1,29 +1,42 @@
...
-  };
+  const [node, setNode] = useState<HTMLElement | null>(null);
+  const observer = useRef<IntersectionObserver>(null);
 
   useEffect(() => {
-    const node = elementRef?.current;
-    const hasIOSupport = !!window.IntersectionObserver;
+    if (node) {
+      if (observer.current) {
+        observer?.current?.disconnect();
```
Aliath: Shouldn't observer be disconnected even if `node` is falsy?

PR #2515, opened 2021-05-03
fix(kebab): postinstall script

amitojsingh366:


amitojsingh366:
something is wrong but im unsure what is, my fix isnt the one tho

Issue #2250, opened 2021-04-22
"enter" button for the chat?


adraneia:
**Is your feature request related to a problem? Please describe.**
ok so i was laying back to my chair, checking up on what i can do on this site, and i thought "Well i could send some emojis" 
so i choose the emojis that i wanted to send and then i realise: i actually have to stop chilling, "get up" and then press the enter(??)

**Describe the solution you'd like**
you could add a button like "chat" or "Send" or an arrow etc, for those that just dont wanna press the enter on their keyboards

**Describe alternatives you've considered**
this is how it looks like rn btw
![send enter](https://user-images.githubusercontent.com/77756019/115711150-dd0f3b80-a37b-11eb-8b80-166549fc2d64.jpg)

Ayushsunny:
Hey @adraneia I wanna work on this issue

goldyydev:
@Ayushsunny here ya go.

Ayushsunny:
> @Ayushsunny here ya go.

Hey @goldyydev could you help me with the codebase please so that I can look for it more easily as I am new to open source

samarmohan:
> Hey @goldyydev could you help me with the codebase please so that I can look for it more easily as I am new to open source

Do you still need help?

Ayushsunny:
> > Hey @goldyydev could you help me with the codebase please so that I can look for it more easily as I am new to open source
> 
> Do you still need help?

yes @samarmohan could help me please that what exactly I need to do here

wes337:
@Ayushsunny 

![fe](https://user-images.githubusercontent.com/9464690/115901112-736d5b00-a469-11eb-84d3-ef987eb73893.PNG)

I believe for this issue, you should add a clickable button here. 
Something like the Material UI "Send" icon (https://material-ui.com/components/material-icons/#material-icons).

So you'll need to:
1. Make an icon component in Kibbeh `src/icons/OutlineSend.tsx`
2. Add it into the `RoomChatInput` component as an icon button which acts as a submit button for the form.

Ayushsunny:
Hello @ajmnz , @goldyydev I have made the changes.
May I make a PR now for this ?

PR #2321, opened 2021-04-25
feat(kibbeh): Added send button to room chat

HenryLeC:
Added a send button to the the room chat input.

Based on idea by @adraneia

closes #2250

PR #2519, opened 2021-05-03
feat(kibbeh): add an enter / send button issue #2250

k-xo:
Add's enter/send button to the chat.

image


dk-raw (approved):

dk-raw:
@ajmnz what do you think of the icon?

amitojsingh366:
we don't want this right now as per @ajmnz 's comment [here](https://github.com/benawad/dogehouse/pull/2321#issuecomment-826390298)

ajmnz:
We probably won't implement this for long (if that even happens).

PR #2518, opened 2021-05-03
Electron header and window changes

BronzW:
- Added minHeight and minWidth to stop people from resizing it so small it's unusable
- Changed the height of the windows minimize icon to prevent it being blurry

PR #2520, opened 2021-05-03
Baklava: Set min width and min height to window

isaxk:
add `minWidth` and `minHeight` properties to the `createMainWindow()` function.

Screenshot 2021-05-03 at 10 04 48
This is what the min-width and height looks like, it is also one pixel above the breakpoint to switch to mobile view.


amitojsingh366:
please use proper commit messages *(`yarn commit` or yarn `commit:signed`)*

dk-raw (approved):

amitojsingh366:
there is already a pr that deals with this and more #2518

PR #2544, opened 2021-05-04
feat(lang): added LOLCAT language

mastrio:
Added new novelty language **'LOLCAT'**, It's based off of the silly Minecraft language LOLCAT (which is
probably based off something else).

*Only 1/3 of the translation is done so far, but I plan to get it done in the next day or two.*

I decided to make the PR now instead of when I finished the translation because I want to get some feedback on it (I'm still not completly happy with the translation)

I mean come one, we need this as a language 😐

😸

### DIFF
--- a/kibbeh/public/locales/en-LOLCAT/translation.json
+++ b/kibbeh/public/locales/en-LOLCAT/translation.json
@@ -1,0 +1,326 @@
+{
+  "_comment": "if you change this file, do: yarn i18",
+  "common": {
+    "loadMore": "mak mor apper",
+    "loading": "lodin........",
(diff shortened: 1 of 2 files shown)


amitojsingh366:
please use proper commit messages *(`yarn commit` or `yarn commit:signed`)*

in your case you have messed up the scope, the scope can only be one of these:
Screen Shot 2021-05-04 at 4 54 45 PM

you can fix this by simple running the following command:
`git commit --amend -m "feat(kibbeh): added LOLCAT language (1/3 translated)"`

mastrio:
huh I thought I did that right, I just realised that I messed up the scope lol.
It should be fixed now 👍

PR #2549, opened 2021-05-04
fix(kibbeh): useResize hook to calculate the estimated size of the message

Gers2017:
Using a custom hook (useResize) to get a size object and using it as a dependecy for the useCallback of the estimateSize

#2356
issss1
isss2
issss3


benawad:
why are you using the y window height?

benawad:
nvm I see the attached issue, interesting fix

PR #2553, opened 2021-05-04
feat: add custom chatThrottle

amitojsingh366:
- kousa: enforces `chatThrottle`
- kibbeh: adds rough UI to customize `chatThrottle`

i added this as per @goldyydev 's request. you should expect a pr from his side soon implementing this into the UI.


overlisted (approved):

dk-raw (approved):

kousa/lib/broth/message/user/get_info.ex
```diff
@@ -43,6 +44,7 @@ defmodule Broth.Message.User.GetInfo do
       field(:numFollowers, :integer)
       field(:online, :boolean)
       field(:lastOnline, :utc_datetime_usec)
+      field(:lastChatMsg, :utc_datetime_usec)
```
benawad: this shouldn't be stored in the db, look at how we throttle in chat genserver

benawad (changes requested):

benawad:
I also don't see any backend tests for the behavior

kousa/lib/broth/message/user/get_info.ex
amitojsingh366: done

amitojsingh366:
> 
> 
> I also don't see any backend tests for the behavior

there are tests that update the `chatThrottle`, check `kousa/test/broth/_message/room/update_test.exs`

Issue #2550, opened 2021-05-04
Duplicated whisper to yourself


Gers2017:
**Describe the bug**
Sending a whisper generate a duplicated message on the chat

**To Reproduce**
1. Go to https://dogehouse.tv
2. Click on "new room"
3. Creta a whisper ``` #@USERNAME ``` for yourself
4. Press enter and enjoy

**Expected behavior**
Only one message please

**Screenshots**
![daaaaa](https://user-images.githubusercontent.com/49139988/117042049-bb417d00-acc8-11eb-8f2d-f8f069a7c44d.png)
![daaaaaaaaaa22222](https://user-images.githubusercontent.com/49139988/117042071-c1cff480-acc8-11eb-8af2-29985e1f5961.png)

![bug123](https://user-images.githubusercontent.com/49139988/117041900-8fbe9280-acc8-11eb-8634-e5c6bbdbc3fa.png)
![bug1234](https://user-images.githubusercontent.com/49139988/117041921-93eab000-acc8-11eb-86db-9836682b8003.png)

**What device are you on?**
Laptop
**Additional context**
```No

```

amitojsingh366:
I can reproduce this

Gers2017:
you mean you can't?

amitojsingh366:
> 
> 
> you mean you can't?

no i said i CAN. i agree with you, this is a thing

PR #2560, opened 2021-05-05
fix(kibbeh): #2550 Fix for duplicated whisper to yourself 

lee1409:
Use set to track message id. Simple fix without affecting others.

Before
image

After

image

### DIFF
--- a/kibbeh/src/modules/room/chat/useRoomChatStore.ts
+++ b/kibbeh/src/modules/room/chat/useRoomChatStore.ts
@@ -65,6 +65,8 @@ export interface RoomChatMessage {
   isWhisper?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


dk-raw (approved):

nadirabbas:
although I never came across this bug, we should actually debug on the backend why whispers are being duplicated.

benawad:
yeah I'm goign to change this on the backend

lee1409:
Yeap. I agree with @benawad and @nadirabbas. At first, I also thought this problem should be fixed on the backend, but wouldn't be great if the frontend can handle the bug in case any duplicated messages may occur in the future? Just my 2 cents. 😄

PR #2578, opened 2021-05-05
feat(baklava): add targz and rpm build

ghost:
This builds a tarball, which is way better than using an AppImage for some. RPMs let Red Hat, openSUSE & Fedora users install DogeHouse like they would as a Debian/Ubuntu user.


TheOtterlord (approved):
Looks good to me, @amitojsingh366?

PR #2589, opened 2021-05-06
fix: remodding a mod won't unmod them in UI

### LINKED ISSUE #2583: Moderator permissions aren't properly parsed by the client
**Describe the bug**
kibbeh's mod_changed event listener simply inverts the current moderator permission that the user has, however, it is possible for this event to be fired on a user who already is a moderator, causing the client not to correctly display the real permission status. This can happen when, for example, a bot (or a client, due to some unexpected behavior) calls `room:set_auth` with level "mod" on a user that is already a moderator.

https://github.com/benawad/dogehouse/blob/43b57e9104ee5998a9182d7dfbee9129db332703/kibbeh/src/shared-hooks/useMainWsHandler.tsx#L248-L251

**To Reproduce**
1. Receive moderator permissions in a room on which you already have them
2. See error

nadirabbas:
when `mod_changed` is dispatched, the UI does `isMod = !isMod`. This is a problem if somebody sends a `room:set_auth = mod`  on an already modded user.

- added tests

fixes #2583

### DIFF
--- a/kibbeh/src/shared-hooks/useMainWsHandler.tsx
+++ b/kibbeh/src/shared-hooks/useMainWsHandler.tsx
@@ -255,7 +255,7 @@ export const useMainWsHandler = () => {
           );
         }
       ),
-      conn.addListener<any>("mod_changed", ({ userId, roomId }) => {
+      conn.addListener<any>("mod_changed", ({ userId, roomId, isMod }) => {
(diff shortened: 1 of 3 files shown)


nadirabbas:
the tests are failing for `kibbeh` for some reason. I think it's when @benawad uncommented the compile errors thingy in `kibbeh`

PR #2607, opened 2021-05-06
(feat): Customize nprogress

sohank1:
nprogress now uses same color as buttons

PR #2611, opened 2021-05-06
feat(kebbeh): change nprogress bar color

sohank1:


sohank1:
I think color is too dark but i was told to use `color-primary-600` or `color-primary-700`

BronzW:
I don't know about other people but I think it would be helpful if you created a new variable named something like progress-bar-color that uses --color-primary-600 and set all the current ones to that variable

sohank1:
> I don't know about other people but I think it would be helpful if you created a new variable named something like progress-bar-color that uses --color-primary-600 and set all the current ones to that variable

What do you mean? Like things that use --color-primary-600 use this new var? I don't think that's a good idea

kibbeh/src/styles/globals.css
```diff
@@ -116,6 +116,19 @@ textarea {
   z-index: 9999999;
 }
 
+#nprogress .bar {
+  background: var(--color-primary-600) !important;
```
amitojsingh366: why this color its barely even visible?

amitojsingh366 (changes requested):

kibbeh/src/styles/globals.css
sohank1: https://github.com/benawad/dogehouse/pull/2607#issuecomment-833746758  I was told to make it this color

sohank1: Changed it and I think it looks better now

sohank1:
@amitojsingh366 Is it good now?
image

amitojsingh366:
> @amitojsingh366 Is it good now?
> image

@ajmnz

benawad:
I'll update the color later if ajb wants a diff one

PR #2630, opened 2021-05-07
dev-env: makes development easier

ghost:
This will make it so I don't have to use vim in my dev container


ghost:
This is the wrong branch

PR #2637, opened 2021-05-07
fix(kousa): add .envrc.bat for Windows.

samarmohan:
Some people are on Windows... Maintainers can change the name of the file if they want.


kousa/.envrc.bat
```diff
@@ -0,0 +1,20 @@
...
+
+set DISCORD_CLIENT_ID=ntahunhaoneuhtohauehonu
+set DISCORD_CLIENT_SECRET=aoneunthaehouh
+
+set GOOGLE_CLIENT_ID=ntahunhaoneuhtohauehonu
+set GOOGLE_CLIENT_SECRET=aoneunthaehouh
+
+set TWITTER_BEARER_TOKEN=aoidjwqdjqiowdjwqadjoqwidoqw
+
+set DATABASE_URL=postgres://asdouaoisdwionasa
```
amitojsingh366: this is not what we want, the env variables shown in CONTRIBUTING.md are just examples and need to be filled manually... we don't want to supply the contributors with our own versions of them.

amitojsingh366 (changes requested):

kousa/.envrc.bat
samarmohan: So what do I add?? Do we want the actual versions, blank versions, or filler ones.

benawad: this is fine, just change this to a legit postgres url pointing at localhost with the database name `kousa_repo2`

benawad (changes requested):

goldyydev:
I think that .envrc and .envrc.bat should be ignored to avoid users leaking their personal info (with shell plugins like direnv who need it to function).

amitojsingh366:
> I think that .envrc and .envrc.bat should be ignored to avoid users leaking their personal info (with shell plugins like direnv who need it to function).

i agree, they should just copy the template from CONTRIBUTING.md into their own .envrc

samarmohan:
Should I just remove .envrc in this PR? I’ll change the goal of it but it will help.

kousa/.envrc.bat
samarmohan: We can just add that in the template env file.

benawad:
I want it to work out of the box with .envrc

PR #2646, opened 2021-05-08
2 new emotes

k0am0:
ggez, wow


benawad:
not taking new emotes atm

Issue #2579, opened 2021-05-05
Hash banned IP addresses


ChrisDesigns:
Much like the idea of #1249, IP addresses should definitively be hashed for privacy reasons. This previous issue describes the GDPR implications of storing plain-text vs hashed identifiers.

In order to handle the changes to the hashing of passwords, the IPs would need to:

1. Create a unique salt (as an unsalted IP can be quickly cracked with a rainbow table)
2. Hashed using said salt
3. Stored into two fields, one the  IP field and the other the IP_Salt (see room_blocks.ex#L47)
4. Insert into the additional key-value database

The process of checking IP addresses then becomes a bit complex, as you don't have the salt on hand. To handle this process, I propose a solution (see image below) using a key–value database, such as Redis, using the IP as the key, to quickly determine if an IP is banned. Otherwise, the server much try each key in the database to determine if a user is banned. This database is best keep in-memory as it's not best to keep unsalted IPs in persistent storage. Meaning that once the  server resets, it will lose this information, but it will still have a the hashed IP version in the long-term storage database.

![Get IP](https://user-images.githubusercontent.com/10255379/117197112-6a5b8280-adb5-11eb-8ff9-ab7149cdb4a0.png)

Alternatively,  a salt can be used for all IP hashing but this still risks a rainbow table attack but gains the benefit of knowing the salt at the time of checking if the IP is banned.

vijairamg:
Is this issue up for grabs?

ChrisDesigns:
> Is this issue up for grabs?

Feel free to create your own branch, I don't have any contribution towards a PR yet. My dev environment isn't set up yet, so I'll probably have small commits this weekend. I'll be willing to collaborate on the work as needed though.

PR #2692, opened 2021-05-10
feat(kousa): implemented ip address hashing, fix tests, etc.

saihnii4:
this pr implements the hashing of ip addresses that are stored in the database and fixes kousa tests.

implements #2579


benawad:
I think I want to use a faster hash than argon

saihnii4:
how about bcrypt?

benawad (changes requested):
switch to HMAC

amitojsingh366:
please don't use npm on kebab, we use yarn for almost all of our workspaces. also delete the `package-lock.json`

kebab/package-lock.json
```diff
@@ -0,0 +1,5716 @@
+{
```
amitojsingh366: please don't use npm on kebab, we use yarn for almost all of our workspaces. also delete the `package-lock.json`

amitojsingh366 (changes requested):

overlisted (commented):
i only review kebab

kousa/Session.vim
```diff
@@ -0,0 +1,118 @@
+let SessionLoad = 1
```
samarmohan: remove this file

samarmohan (approved):

kibbeh/package.json
```diff
@@ -30,6 +30,7 @@
     "@material-ui/pickers": "^3.3.10",
     "@svgr/webpack": "^5.5.0",
     "@tailwindcss/line-clamp": "^0.2.0",
+    "bufferutil": "^4.0.3",
```
benawad: remove unrelated code

kousa/config/dev.exs
```diff
@@ -31,6 +31,8 @@ config :kousa,
   api_url: System.get_env("API_URL") || "http://localhost:4001",
   secret_key_base:
     "213lo12j3kl21j3kl21alaksjdklasjdklajsldjsaldjlasdlaksjdklasjdklajsldjsaldjlasdlaksjdklasjdklajsldjsaldjlasdadjlasjddlkijoqwijdoqwjd12loki3jhl12jelk12jekl1221099dj120",
+  ip_hashing_key:
```
benawad: remove this

kousa/lib/kousa/utils/hash.ex
```diff
@@ -1,21 +1,29 @@
...
 
   @spec hash_ip(String.t()) :: hash_result
   def hash_ip(ip) do
-    Argon2.hash_pwd_salt(ip)
+    if is_nil(ip) do
+      {:error, "ip is nil"}
+    else
+      # Should we refactor this so we don't pipe a function call?
+      # Encodes Hash into Base16
+      :crypto.mac(:hmac, :sha256, Application.fetch_env!(:kousa, :ip_hashing_key_base), ip)
```
ChrisDesigns: This isn't a good implementation.  It would be very easy to get the base key and create a rainbow table for the whole IP space. This methodology uses ip_hashing_key_base as "pepper".  To do this method properly, it should be done in conjunction with salts.  Looking at [RFC7518](https://datatracker.ietf.org/doc/html/rfc7518#section-3.2), the uniquely created salt should two time the length of the input (this would  mean 2*45 = 90 characters). The hash can be stored in the same field as the IP using the [crypt](https://en.wikipedia.org/wiki/Crypt_(C)) format. A new function could be created to handle the splitting of these two fields and use secure_compare to return if the IP + salt = hash

willdoescode (changes requested):
Remove Session.vim file

samarmohan:
> Remove Session.vim file

Lol I already told him to do that

kousa/lib/kousa/utils/hash.ex
benawad: I thought the whole point is someone wouldn't have the hash key because its a secret

kousa/Session.vim
saihnii4: done

kousa/lib/kousa/utils/hash.ex
ChrisDesigns: The key should not be a substitute for a good salt. I think that this approach shares the same issues as a shared [pepper](https://en.wikipedia.org/wiki/Pepper_(cryptography)#Shared_Secret_Pepper).
Looking at it from the prospective of a hacker or an evil company, this value is store via environment variables `System.get_env("IP_HASHING_KEY_BASE")`. Being that value is constant (shared for all hashing) and can be fetched by both a company or hacker if they gain access to the system. It won't take much more work for either entity to create a rainbow table using this know secret (presuming they managed to get database access). Every IP is bijective with the entries of a rainbow table for a know key. The number of IP that exists in the database, denoted as **_n_**, is only a subset of the know rainbow table entity. So let's take all the Class C IPs, denoted as **_c_**, to create a valid rainbow table, the IP space would need to be computed for each salt. This will take the complexity from O(**_c_**) to O(**_c_*****_n_**). Don't get me wrong, I think having anything is better than no hashing. I just think the approach could be broken based on  the methods displayed.

saihnii4: how about generating and encrypting a secret private key using a encryption key and storing that private key in the database alongside the user? this method forces an attacker to get access to the database **and** the environment variables.

saihnii4:
closing because stale ig

PR #2699, opened 2021-05-11
refactor(dolma): use switch instead of if in filterUnitoken.ts

MeztliRA:


iamanalt:
... why?

HoloPanio:
Sorry but we're not going to be using switches, we will not have the values hard coded for much longer so we do not need to add a switch

amitojsingh366:
@HoloPanio the creator of `dolma` doesn't want switch statements so im gonna close this pr

PR #2717, opened 2021-05-12
fix(kibbeh): implements fix suggested by #2689 : emote autocomplete

### LINKED ISSUE #2689: twitch emotes autocomplete
when you wanna write a emote in chat like for example :5Head: 

I would write :5h and auto-complete but instead of deleting the :5h it would just add the completed one 

like this 
image

it would be really great and time efficient 
thanks

vijairamg:
fix #2689


benawad:
I don't understand, why is this a desired behavior? For emote spamming? Discord/Twitch don't do it.

cc @MrBaggieBug

ghost:
I believe the better behavior would be to add a thing that pops up when you're typing an emoji name, and allowing you to press tab to put it in. Discord does this. Not sure about Twitch though.

PR #2726, opened 2021-05-12
fix(global): update npm deployment scripts

HoloPanio:


.github/workflows/globalkey-npm_deploy.yaml
```diff
@@ -17,22 +16,31 @@ jobs:
...
           node-version: 14
+      - name: Yarn cache
+        uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: |
+            .yarn/cache
+          key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
```
amitojsingh366: no need to add this here this package has no node deps, it's written in rust

amitojsingh366 (changes requested):

PR #2702, opened 2021-05-11
feat: Scheduled rooms tab on user profile

amitojsingh366:
- kibbeh: added scheduled rooms tab on user profile
- kousa: make `room:get_scheduled` send anyone's scheduled rooms if provided with a valid `userId`
- kousa: fixed tests for `user:admin_update` *(cuz the checks were failing)*

PR #2728, opened 2021-05-12
feat: add admin tab to user profile

amitojsingh366:
as per @benawad 's request
this pr depends on #2702 so merge that first, the diff should go down after that's done


overlisted (approved):

kousa/lib/kousa/user.ex
```diff
@@ -79,7 +79,7 @@ defmodule Kousa.User do
   def admin_update_with(changeset, admin) do
     authorized_github_id = Application.get_env(:kousa, :ben_github_id, "")
 
-    with %{githubId: ^authorized_github_id} <- admin do
+    if admin.staff == true or admin.githubId == authorized_github_id do
```
benawad, line 82: add a test for this logic

benawad (changes requested):

kousa/lib/kousa/user.ex
amitojsingh366: done

PR #2730, opened 2021-05-13
CONTRIBUTING.md changes 

phocodes:
I am making it clear for new contributors that it would be appreciated if they used languages that were mainly used to make DogeHouse


imadill:
Why bother writing these when there is already wrote in the README which directory contains what.

Plus there are some typos. Also we're not using just HTML and Vanilla JS, we're using Next.js.
Also we're neither using Java nor Objective-C to build the apps, we're using React Native.

AmaaniGoose:
First time I'm seeing a text only PR that manages to be flat out wrong

TheOtterlord:
The README explains what projects do what, and each project explains what language and framework(s) it uses.

PR #2734, opened 2021-05-13
fix(kibbeh): upcomingroomscard.tsx today "k:mm a" > "h:mm a" 

NickersWeb:


ajmnz:
Why change `00:00` to `24:00`?

NickersWeb:
This solves the issue of showing "0:00pm" when set to "12:00pm" in a scheduled room. https://date-fns.org/v2.21.3/docs/format I'm not sure why "k" format was showing "0" when the documents say otherwise, but it seems to solve it locally.

ajmnz:
> This solves the issue of showing "0:00pm" when set to "12:00pm" in a scheduled room. https://date-fns.org/v2.21.3/docs/format I'm not sure why "k" format was showing "0" when the documents say otherwise, but it seems to solve it locally.

Oh makes sense then

Issue #2733, opened 2021-05-13
Search: show no results message when there only 2 characters typed


ghost:
## Searchbar Message
Currently it takes at least more than 2 characters in the searchbar to trigger the actual search. If someone doesn't type more than 2 characters, they'l get an blank search result box. It would be nice to get a hint or the "no results" message :)

![Bildschirmfoto 2021-05-13 um 11 45 58](https://user-images.githubusercontent.com/10587159/118109204-2d107980-b3e1-11eb-9103-e6bf77622e16.png)

XfaonXE:
I think thats something with SQL, or maybe its just coded like that

PR #2756, opened 2021-05-14
feat(kibbeh): add loader to search bar

### LINKED ISSUE #2741: add a loader when searching for stuff
sometimes it takes a little while to load the results when searching stuff on dogehouse. A loader would be nice.
image

akashrajum7:
Added loading to search bar that shows when search is making request to the server and show `no results` when no results are present

fix #2741 and #2733


akashrajum7:
Looks like a type issue, looking into it.

samarmohan:
Why is this a breaking change?

kibbeh/src/icons/LoadingIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+      fill="none"
+      viewBox="0 0 24 24"
+    >
+      <circle
+        className="opacity-25"
+        cx="12"
+        cy="12"
+        r="10"
+        stroke="currentColor"
+        stroke-width="4"
```
samarmohan: should be strokeWidth. that will fix the failing checks.

akashrajum7:
I read in the docs that changing props is a breaking change, sorry if I misunderstood.

kibbeh/src/modules/search/SearchBarController.tsx
```diff
@@ -90,6 +90,7 @@ export const SearchBarController: React.FC<SearchControllerProps> = ({}) => {
                 ? t("components.search.placeholderShort")
                 : t("components.search.placeholder")
             }
+            isLoading={isLoading}
```
samarmohan, line 93: I believe you can just use ES6 and write `isLoading` instead of ` isLoading={isLoading}`

kibbeh/src/modules/search/SearchPage.tsx
```diff
@@ -32,10 +33,15 @@ export const SearchPage: PageComponent<LoungePageProps> = ({}) => {
...
-              conn.query.search(e.target.value).then((r) => setResults(r));
+              setSearchLoading(true);
+              conn.query.search(e.target.value).then((r) => {
+                setResults(r);
+                setSearchLoading(false);
+              });
             }}
             searchPlaceholder="Search"
             onBackClick={() => router.back()}
+            searchLoading={searchLoading}
```
samarmohan, line 43: use ES6 here

samarmohan (changes requested):

samarmohan:
> I read in the docs that changing props is a breaking change, sorry if I misunderstood.

@amitojsingh366 just confirmed it is not.
Can you update the description?

akashrajum7:
Thanks a lot for the help @samarmohan, will work on the requested changes.

samarmohan:
> Thanks a lot for the help @samarmohan, will work on the requested changes.

Welcome!

kibbeh/src/modules/search/SearchBarController.tsx
akashrajum7: Doing that is making isLoading true always, I think we can only do that on objects.

kibbeh/src/modules/search/SearchPage.tsx
akashrajum7: Same reason as mentioned above, can only be done on objects.

kibbeh/src/icons/LoadingIcon.tsx
akashrajum7: Done thanks.

akashrajum7:
@samarmohan I have done the fixes, doing `yarn format` is formatting a lot of files so didn't do it, should I do it in a separate pull request?

samarmohan:
Don't worry about formatting. I'll get someone to merge this ASAP!

ajmnz (changes requested):
Remove the icon you added and change it for the one all other components use:

https://github.com/benawad/dogehouse/blob/staging/kibbeh/src/ui/Button.tsx#L65

akashrajum7:
Done @ajmnz

amitojsingh366:
image
why does the spinner look so small?

ghost:
huh? that looks normal imo

amitojsingh366:
and why does it only show after typing more than 2 words?

akashrajum7:
@amitojsingh366 It's the same size as the search icon, 16px.

akashrajum7:
@amitojsingh366 Because it only shows up when the API call is being made I.e, search is happening.

amitojsingh366:
> 
> 
> @amitojsingh366 Because it only shows up when the API call is being made I.e, search is happening.

try to make it show whenever there are no results

akashrajum7:
> try to make it show whenever there are no results

I don't think that's appropriate UX as the user might keep waiting for the result to show up.

akashrajum7:
I will work on a separate pull request that will show `no results` when there are no results, so that shouldn't be a problem.

akashrajum7:
Or shall I make those changes to this pull request itself? It will solve https://github.com/benawad/dogehouse/issues/2733, I have already worked on it, just need to make the pull request.

Update: I will merge that change also into this since both belong to search, I think it makes sense.

ajmnz:
Size and behavior are both fine

ajmnz (approved):

PR #2758, opened 2021-05-14
fix(kibbeh): updated text parser regex

### LINKED ISSUE #2725: Can't load profile in Safari
**Describe the bug**
I wanted to change my name, so I clicked on "Profile" from the top right menu. The spinner in the top right spun forever and the page never loaded.

**To Reproduce**
1. Go to the Home Screen
2. Click on your avatar
3. Click on 'profile'

**Expected behavior**
Profile loads

**Screenshots**
image

**What device are you on?**
Macbook, Safari 14.0.3

**Additional context**
New account, created from Github within the last 15 minutes

mchahed99:
Safari doesn't support lookbehind.

Fix #2725

### DIFF
--- a/kibbeh/src/modules/display/TextParser.tsx
+++ b/kibbeh/src/modules/display/TextParser.tsx
@@ -9,9 +9,9 @@ interface TextParserProps {
 
 export const TextParser: React.FC<TextParserProps> = ({ children }) => {
   return (<>
-    {children.split(/(?=[ ,\n])|(?<=[ ,\n])/g).map((text, i) => {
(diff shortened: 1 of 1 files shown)


samarmohan (approved):

HoloPanio:
This is unnecessary because it is handled by dolma.

n3oney:
@HoloPanio are you sure about that?
Screenshot 2021-05-17 at 10 23 34
It seems to be the reason why the profile page won't load on Safari.
Once that fix is applied it loads fine.
Screenshot 2021-05-17 at 10 24 19

PR #2797, opened 2021-05-17
fix(kibbeh): fix header clip on iOS with keyboard open

### LINKED ISSUE #2496: Header clipped on ios safari
**Describe the bug**
image

**To Reproduce**
Keyboard is up, header disappears.

**Expected behavior**
Header should be visible without dismissing

**What device are you on?**
iPhone SE. ios chrome (webkit)

**Additional context**
Can be fixed by detecting keyboard is up.

https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API

n3oney:
fix #2496
this also creates a new hook similar to `useWindowSize()` that uses the [VisualViewport API](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport) instead.
That hook will use the window size if the VisualViewport API is not supported (Firefox)


n3oney:
image

dk-raw (approved):

PR #2800, opened 2021-05-17
fix(kibbeh): update OWO Translation

jaipack17:
Edited some phrases, spellings and more. Added some new emojis and fixed capitalizations. 

Thanks, 
jaipack17


willdoescode:
Very needed fix

dk-raw:
much true

PR #2820, opened 2021-05-18
fix(kibbeh): update Sanskirt Translation

0xEquinox:
Partial translation.


willdoescode:
I am will and I approve this message

PR #2825, opened 2021-05-18
fix(dolma): make emotes case-insensitive

prokopschield:


jaipack17:
Hmm as far as I know. Emotes shouldnt be case sensitive. But can be used when sending emojis. For example - you type in hello and :DogeHello: shows up.

PR #2829, opened 2021-05-19
fix(kibbeh): redirect with 301 instead of 404

### LINKED ISSUE #2811: Remove the 404 page and use the Redirects feature instead
```dogehouse/kibbeh/src/pages/404.tsx``` is redundant since we don't have a 404 page
 it is a simple redirect. We should instead use the [Redirects](https://nextjs.org/docs/api-reference/next.config.js/redirects) feature built into the Next by modifying 
```next.config.js``` This should significantly speed up redirects and send a 301 redirect response code to browsers.

```js
module.exports = {
  ...,
  async redirects() {
    return [
      {
        source: '/404',
        destination: '/',
        permanent: true,
      },
    ]
  },
}
```

rocktimsaikia:
fix #2811

The suggestion to completely remove the `404 page` and use the `next redirect`  does not actually work. Next.js does not let us grab the 404 redirects like that. There is an [open issue](https://github.com/vercel/next.js/discussions/16749) on that with a workaround which is currently being used in the Dogehouse frontend too.  So that basically ends that primary point on that issue. 

But the use of `301 redirects` instead of `404` is reasonable. But we can not write the status code with `next router`. So I ended up using [**nextjs-redirect**](https://www.npmjs.com/package/nextjs-redirect) which is a zero dependency module that solves this issue with no hassle. By default, it sets the `301` status code so there is no config needed on that.

Another reason to use `301` instead of `404` is that in browsers like Brave it detects the 404 redirects and shows a dialogue box on top of the screen. Using `301` solves that issues.

### DIFF
--- a/kibbeh/package.json
+++ b/kibbeh/package.json
@@ -44,6 +44,7 @@
     "lodash": "^4.17.21",
     "mediasoup-client": "^3.6.30",
     "next": "10.2.0",
+    "nextjs-redirect": "^5.0.2",
     "normalize-url": "4",
     "nprogress": "^0.2.0",
     "rc-slider": "^9.7.2",
--- a/kibbeh/src/pages/404.tsx
+++ b/kibbeh/src/pages/404.tsx
@@ -1,10 +1,7 @@
-import { useEffect } from "react";
-import { useRouter } from "next/router";
+import redirect from "nextjs-redirect";
+
+const Redirect = redirect("/"); // By default, it sends a 301 status code so not need to specify
 
 export default function Custom404() {
-  const router = useRouter();
-  useEffect(() => {
-    router.replace("/dash");
-  });
-  return null;
+  return <Redirect />;
 }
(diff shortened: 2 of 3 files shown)


amitojsingh366:
we don't need an external package for this, and why have you deleted` yarn.lock`. i'm gonna close this pr as it is unnecessary

rocktimsaikia:
@amitojsingh366 How do you suggest redirecting with a `301` status code? It's a zero dependency package so I don't think it would have made any performance or build size issue. And I didn't delete the `lock` I think it was updated because of the newly added package.

Anyway, this issue remains in Brave browser because of the `404` status. Also `404` redirect is misleading since the goal is to not have any 404 pages.

issue

amitojsingh366:
> @amitojsingh366 How do you suggest redirecting with a `301` status code? It's a zero dependency package so I don't think it would have made any performance or build size issue. And I didn't delete the `lock` I think it was updated because of the newly added package.
> 
> Anyway, this issue remains in Brave browser because of the `404` status. Also `404` redirect is misleading since the goal is to not have any 404 pages.
> 
> issue

ill handle it

PR #2855, opened 2021-05-21
fix(profileheader): append username if displayName is not set

### LINKED ISSUE #226: Users without a display name appear blank
If someone with a GitHub account without a display name joins the room, their name in the grid appears blank instead of using their username.

name in listeners list: 
image

name in profile window:
image

byeze:
when people joins DogeHouse with an account that doesn't have a display name, ProfileHeader doesnt
show any displayname. So, I appended the username of the user when displayname is not set

fix #226

**BY ERROR:** Added another commit of some translated strings

### DIFF
--- a/kibbeh/public/locales/es/translation.json
+++ b/kibbeh/public/locales/es/translation.json
@@ -15,8 +15,8 @@
     "copied": "¡Copiado!",
     "formattedIntlDate": "{{date, intlDate}}",
     "formattedIntlTime": "{{time, intlTime}}",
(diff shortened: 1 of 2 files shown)


amitojsingh366:
please follow the proper pattern for commit messages *(💡TIP: you can use `yarn commit` or `yarn commit:signed` and answer the questions asked to generate a properly formatted one for you!)*

to fix your most recent commit you can run the command `git commit --amend -m "properly formatted commit message here"`

to fix your older/multiple commits please take a look at [this article](https://docs.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message#amending-older-or-multiple-commit-messages)

dk-raw (approved):

PR #2856, opened 2021-05-22
New rewritten client for kebab

overlisted:
- Audio stuff didn't get refactored because it still uses the old request format
- Kept the old wrapper because kibbeh uses it
- Only made one example use it because ummm


ajmnz (approved):
Besides linting errors, lgtm

